PR #6, opened 2019-02-05
Fix develop

iamdanfox:
We didn't have branch protection set up initially, but we do use bulldozer in blacklist mode, so a few PRs merged with red builds.


ferozco:
👍

PR #7, opened 2019-02-06
Add palantir.bintray.com repo to example

dansanduleac:


dansanduleac:
tbh we should just publish to the gradle portal instead... but this is fixing the currently broken example.

iamdanfox (approved):

PR #8, opened 2019-02-06
[feature] New `getVersion(String, Configuration)` function

iamdanfox:
## Before this PR

People are using nebula's `dependencyRecommendations.getRecommendedVersion` thing all over the place

## After this PR

We make an acceptable workaround available, and recommend people delete all other usages of that function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,3 +263,26 @@ This plugin requires the settings.gradle to declare a `rootProject.name` which i
  include 'sls-status-api:sls-status-api-objects'
  include 'sls-status-dropwizard'
 ```
+
+#### What about `dependencyRecommendations.getRecommendedVersion`?
+
+If you rely on this Nebula function, then gradle-consistent-versions has a similar alternative:
+
+```diff
+-println dependencyRecommendations.getRecommendedVersion('com.google.guava:guava')
++println getVersion('com.google.guava:guava')
+```
+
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -263,3 +263,19 @@ This plugin requires the settings.gradle to declare a `rootProject.name` which i
...
+```diff
+-println dependencyRecommendations.getRecommendedVersion('com.google.guava:guava')
++println getVersion('com.google.guava:guava')
+```
+
+Note that you can't invoke this function at configuration time (e.g. in the body of a task declaration), because the plugin needs to resolve dependencies to return the answer and Gradle [strongly discourages](https://guides.gradle.org/performance/#don_t_resolve_dependencies_at_configuration_time) resolving dependencies at configuration time.
+
+Alternatives:
+
+- if you rely on this function for sls-packaging `productDependencies`, use `detectConstraints = true` instead
```
dansanduleac: or upgrade to sls-packaging 3.x

README.md
```diff
@@ -263,3 +263,19 @@ This plugin requires the settings.gradle to declare a `rootProject.name` which i
...
+- if you rely on this function to configure the `from` or `to` parameters of a `Copy` task, consider moving the copy action into a  doLast block.
```
dansanduleac: they can use a closure

```gradle
to { "$buildDir/dest/${getVersion('foo:bar')}/" }
```

dansanduleac (approved):

PR #20, opened 2019-02-19
[fix] Write locks could fail under certain conditions

dansanduleac:
## Before this PR

Write locks could fail like so, if projects end up being evaluated in a particularly pathological order and some other conditions are true:

```
* What went wrong:
A problem occurred evaluating project ':foo:bar'.
> Failed to apply plugin [class 'org.gradle.api.plugins.BasePlugin']
   > DefaultDomainObjectSet#add(T) on [DefaultProjectDependency{dependencyProject='root project 'does-not-fail-if-subproject-evaluated-later-applies-base-plugin-in-own-build-file'', configuration='subprojectUnifiedClasspathCopy'}, DefaultProjectDependency{dependencyProject='project ':foo'', configuration='subprojectUnifiedClasspathCopy'}] cannot be executed in the current context.
```

Specifically, if
* subproject A depends on subproject B, 
* B applies the `base` plugin in its `build.gradle` (as opposed to the root project's build file configuring `project('b') { apply plugin: 'base' }`), 
* A is configured before B when calling `rootProject.allprojects`

Then you'd get the above error.

## After this PR

Fixed that failure case, and added a test to guard against it.

### DIFF
--- a/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
@@ -244,20 +244,24 @@ public class VersionsLockPlugin implements Plugin<Project> {
                     ResolveConfigurationsTask.class, task -> task.mustRunAfter(":resolveConfigurations"));
         }
 
+        NamedDomainObjectProvider<Configuration> subprojectUnifiedClasspath =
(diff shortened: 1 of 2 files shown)


iamdanfox:
:+1: awesome bug hunting

iamdanfox:
@pkoenig10 have just tagged 0.16.0, which should fix the case you encountered yesterday: https://github.com/palantir/gradle-consistent-versions/releases/tag/0.16.0

PR #23, opened 2019-02-22
[feature] publish-bom plugin

dansanduleac:
## Before this PR

Currently we publish BOMs with all of a project's dependencies, based on nebula.dependency-recommender's functionality:
```gradle
nebulaDependencyManagement.fromConfigurations { configurations.runtimeClasspath }
```

However, in the absence of nebula.dependency-recommender, it's difficult to produce a similar BOM.

## After this PR

A new plugin is introduced, `com.palantir.publish-bom`, that publishes a BOM containing 
* _all_ of the constraints from the lock file
* constraints for each other project in the build that has `publish-maven` applied and at least one publication
* maven imports for any platforms used in `rootConfiguration`

Applying it requires that `com.palantir.consistent-versions` is applied on the root project, or else an error is thrown.

TODO:
- [ ] how it works


src/main/java/com/palantir/gradle/versions/ConsistentPlatformPlugin.java
```diff
@@ -0,0 +1,191 @@
...
+
+        project
+                .getPlugins()
+                .withType(JavaBasePlugin.class)
+                .configureEach(plugin -> failBecauseJavaPluginApplied(project));
+    }
+
+    private static void failBecauseJavaPluginApplied(Project project) {
+        throw new GradleException("Cannot apply " + ConsistentPlatformPlugin.class + " to project that has the java "
+                + "plugin applied: " + project);
```
iamdanfox: error msg to include clear action pls?

iamdanfox:
Closing out stale PRs

PR #27, opened 2019-02-26
[improvement] getVersion(String, String) signature allows easier adoption

iamdanfox:
## Before this PR

It was tricky to excavate out the change from nebula -> gradle-consistent-versions because of the `dependencyRecommendations.getRecommendedVersion` function. In nebula it took two args, but in our plugin it took one arg.

## After this PR

our plugin now supports the two-arg signature, so we can excavate people using a simple sed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,7 +285,7 @@ Alternatives:
 
 ```diff
  task copySomething(type: Copy) {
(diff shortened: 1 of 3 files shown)


ferozco:
👍

PR #40, opened 2019-03-05
[feature] new ./gradlew checkMinimalVersions task

iamdanfox:
## Before this PR

People often had lots of cruft in their versions.props.

## After this PR

We now have a helpful task to delete some of this.

TODO:

- [ ] this logic doesn't actually get you a minimal result


iamdanfox:
Closing out for now as deprioritised

PR #42, opened 2019-03-06
[improvement] Lock platforms too

dansanduleac:
## Before this PR

We were explicitly trying to filter out platforms from the lock file, based on the initial assumption that virtual platforms would show up as platforms in the resolution result.
That has been fixed in gradle 5.1, so now that was only filtering out real published platforms (e.g. boms). It's actually correct to include these platforms and all their dependents in the lock file, as they are real dependencies.

## After this PR

We now include platforms in the lock file.
For users using platform dependencies, a `./gradlew --write-locks` will be necessary in order for their lock file to pick up the platforms that were previously being filtered out.


iamdanfox:
Let's write a test to capture this behaviour :+1:

PR #43, opened 2019-03-06
Use local maven repo in integration test rather than jcenter

dansanduleac:
## Before this PR

Integration tests were being run against real external repos.

## After this PR

They are now run against internally-generated repos with fake jars, since all we care about is their inter-dependencies.


iamdanfox:
This is really slick! :+1:

dansanduleac:
argh didn't finish :(

PR #45, opened 2019-03-06
Use local maven repo in integration test rather than jcenter [FLUP]

dansanduleac:
Same idea as #43 but for VersionsPropsPluginIntegrationSpec which hadn't been migrated yet when that merged

### DIFF
--- a/src/test/groovy/com/palantir/gradle/versions/VersionsPropsPluginIntegrationSpec.groovy
+++ b/src/test/groovy/com/palantir/gradle/versions/VersionsPropsPluginIntegrationSpec.groovy
@@ -18,19 +18,29 @@ package com.palantir.gradle.versions
 
 import groovy.util.slurpersupport.GPathResult
 import groovy.util.slurpersupport.NodeChildren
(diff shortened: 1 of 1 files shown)


src/test/groovy/com/palantir/gradle/versions/VersionsPropsPluginIntegrationSpec.groovy
```diff
@@ -18,19 +18,29 @@ package com.palantir.gradle.versions
 
 import groovy.util.slurpersupport.GPathResult
 import groovy.util.slurpersupport.NodeChildren
-import nebula.test.IntegrationTestKitSpec
 
-class VersionsPropsPluginIntegrationSpec extends IntegrationTestKitSpec {
+class VersionsPropsPluginIntegrationSpec extends IntegrationSpec {
```
iamdanfox, line 22: still really unhappy that we're not using nebula-test as IntelliJ test running consistently uses the wrong code (cc @ferozco)

iamdanfox:
:+1:

PR #66, opened 2019-03-24
[improvement] Set up project for publishing to Gradle plugins repo

markelliot:
## Before this PR

consistent-versions plugin is not available on Gradle plugins portal

## After this PR

plugin will become available on Gradle plugins portal after next publish


ferozco:
change looks good. @dansanduleac any reason we hadn't done this earlier?

build.gradle
```diff
@@ -76,3 +79,7 @@ gradlePlugin {
         }
     }
 }
+
+publishPlugins.onlyIf {
+    project.version ==~ /\d+\.\d+\.\d+/
```
iamdanfox: would prefer to use `versionDetails().isCleanTag` instead of duplicating regexes.

dansanduleac:
@markelliot you'll want to change `publish-jar.gradle` like this: https://github.com/palantir/sls-packaging/pull/438/files#diff-a473ae678476f9680c743c276c83379bR31, instead of using `nebula.maven-resolved-dependencies`, as that one is incompatible with the plugin-publish plugin. As in, `plugin-publish` has its own bespoke logic for creating the POM, and it's not hookable.
The logic linked above makes gradle inject the resolved versions in a native way (this functionality has been introduced in gradle 5.2).

build.gradle
```diff
@@ -57,10 +59,11 @@ tasks.withType(JavaCompile) {
...
-gradlePlugin {
-    // do not add new task to publish to plugins.gradle.org
-    automatedPublishing = false
-
+pluginBundle {
+    website = 'https://github.com/palantir/gradle-consistent-versions'
+    vcsUrl = 'https://github.com/palantir/gradle-consistent-versions'
+    description = 'Compact, constraint-friendly lockfiles for your dependencies'
+    tags = ['versions']
     plugins {
```
dansanduleac, line 63: Think you want to keep this block under `gradlePlugin` as it's a bit different (it configures the `java-gradle-plugin` plugin that generates the plugin descriptor files).
`pluginBundle.plugins` will actually be auto-populated from the ids defined in `gradlePlugin.plugins`, see the [example](https://plugins.gradle.org/docs/publish-plugin). Although you can still configure `pluginBundle.plugins` separately if you want to give them display names, though I don't think it's necessary.

dansanduleac:
Remaining work here:
Fix https://github.com/nebula-plugins/nebula-publishing-plugin/blob/master/src/main/groovy/nebula/plugin/publishing/maven/MavenCompileOnlyPlugin.groovy so that it doesn't add a new dependencies section, as this breaks the POM.

One possible solution: apply it (it's applied through `nebula.maven-publish`) _after_ the `plugin-publish` plugin, since the former will detect an already existing dependencies block in the BOM and thus not corrupt it, whereas `plugin-publish` assumes there isn't a dependencies section.

Another: make `nebula.maven-compile-only` only modify publications if they have the java component.

iamdanfox:
Let's just do the same thing as https://github.com/palantir/sls-packaging/blob/develop/gradle-sls-packaging/build.gradle#L26-L46

dansanduleac:
👍

iamdanfox:
Let's push some RC tag on this branch so that we can verify it's actually usable before merging to develop!

iamdanfox:
No luck so far? https://plugins.gradle.org/search?term=com.palantir.consistent-versions

dansanduleac:
Nope.. the tags I cut seem to have tried to publish the wrong version somehow! (so bintrayUpload fails before the plugin gets to be published)

I tried this from ssh as well but it seems like it didn't resolve the versions properly, which was unexpected given our change to `publish-jar.gradle`...

```
Caused by: com.gradle.publish.MissingDependencyVersionException: No version found for com.palantir.configurationresolver:gradle-configuration-resolver-plugin on pom generation.
	at com.gradle.publish.PomWriter.addDependency(PomWriter.java:58)
	at com.gradle.publish.PomWriter.createPomDocument(PomWriter.java:44)
	at com.gradle.publish.PomWriter.writePom(PomWriter.java:30)
	at com.gradle.publish.PublishTask.generatePom(PublishTask.java:168)
	at com.gradle.publish.PublishTask.ensurePomIsAvailable(PublishTask.java:154)
	at com.gradle.publish.PublishTask.publish(PublishTask.java:70)
```

dansanduleac:
Published successfully via SSH
```
> Task :publishPlugins
Publishing plugin com.palantir.consistent-versions version 1.3.2-rc1
Thank you. Your new plugin com.palantir.consistent-versions has been submitted for approval by Gradle engineers. The request should be processed within the next 24 hours, at which point you will be contacted via email.
Publishing artifact build/libs/gradle-consistent-versions-1.3.2-rc1.jar
Publishing artifact build/libs/gradle-consistent-versions-1.3.2-rc1-sources.jar
Publishing artifact build/libs/gradle-consistent-versions-1.3.2-rc1-javadoc.jar
Publishing artifact build/libs/gradle-consistent-versions-1.3.2-rc1-groovydoc.jar
Publishing artifact build/publications/pluginMaven/pom-default.xml
Activating plugin com.palantir.consistent-versions version 1.3.2-rc1
```

dansanduleac:
👍

ferozco:
👍

PR #67, opened 2019-03-25
Test that virtual platform is applied across projects

dansanduleac:
Add a test that ensures a virtual platform from `versions.props` aligns components coming from different projects.


iamdanfox:
:+1:

PR #75, opened 2019-03-27
[improvement] Fail the build if `versions.lock` is missing

dansanduleac:
## Before this PR

The plugin would tolerate a missing `versions.lock` with a warning.

## After this PR
It now fails when `versions.lock` is missing.

It's still possible to allow the build to ignore the lock file by running `./gradlew -PignoreLockFile`, which is only intended for debugging purposes.


iamdanfox:
:+1:

PR #95, opened 2019-04-05
[improvement] Don't apply / depend on configuration-resolver

### LINKED ISSUE #94: Do we really need to transitively apply configuration-resolver plugin?
## What happened?

I tried to use this plugin on a clean repo:

```
plugins {
    id  'com.palantir.consistent-versions' version '1.3.2-rc3'
}
```

But got this error:

```
* What went wrong:
A problem occurred configuring root project 'foo'.
> Could not resolve all artifacts for configuration ':classpath'.
   > Could not find com.palantir.configurationresolver:gradle-configuration-resolver-plugin:0.3.0.
     Searched in the following locations:
       - https://plugins.gradle.org/m2/com/palantir/configurationresolver/gradle-configuration-resolver-plugin/0.3.0/gradle-configuration-resolver-plugin-0.3.0.pom
     Required by:
         project : > com.palantir.consistent-versions:com.palantir.consistent-versions.gradle.plugin:1.3.2-rc3 > com.palantir.gradle.consistentversions:gradle-consistent-versions:1.3.2-rc3
```

No failures pls

dansanduleac:
## Before this PR

We depend on `com.palantir.configurationresolver:gradle-configuration-resolver-plugin` which isn't published to the gradle plugin portal.
Therefore, if someone attempts to depend on us via the plugins DSL, without specifying palantir's bintray as an additional buildscript repository, e.g.
```
plugins {
	id 'com.palantir.consistent-versions' version '1.3.2-rc3'
}
```

Then their resolution will break.

One solution would be to depend on the coordinate that *is* published, `gradle.plugin.com.palantir.configurationresolver:gradle-configuration-resolver-plugin`, but we don't want to continue linking these two plugins unnecessarily.

## After this PR
==COMMIT_MSG==
Since `com.palantir.configuration-resolver` is not necessary to this plugin, stop depending on it.
==COMMIT_MSG==

fixes https://github.com/palantir/gradle-consistent-versions/issues/94

## Possible downsides?
Users relying on us applying this plugin will break.
The remediation should be easy though: just apply the plugin themselves.
All palantir repositories apply `com.palantir.configuration-resolver` already.


iamdanfox:
:+1: Longer-term, I'd like to completely get rid of configuration-resolver!!

PR #122, opened 2019-05-15
[fix] Resolve lock state the same whether writing locks or not

dansanduleac:
## Before this PR

When running with `--write-locks`, we transitively copy all dependencies of `unifiedClasspath` through project dependencies, and then resolve this copied graph. This is so we don't force the actual java configurations in subprojects, so that we can still apply the locked constraints afterwards.

When running without, however, we don't do the above, and just resolve `unifiedClasspath` directly, in order to determine the "current lock state".
But this means that existing constraints from the lock file would feed back into the current lock state, which might cause it to be different from what you'd get if you ran with `--write-locks`.

## After this PR
==COMMIT_MSG==
Resolve the lock state in the same way whether running with `--write-locks` or not, to make sure they are consistent.
==COMMIT_MSG==

## Possible downsides?
* If the resolution we do now (via transitively copying configurations) is broken, for instance if it ends up missing dependencies we are actually using, then we end up just not locking those dependencies, and we wouldn't know about it.


iamdanfox (approved):

dansanduleac:
👍

PR #123, opened 2019-05-15
[improvement] Move lock verification to a task that check depends on

dansanduleac:
## Before this PR

We only verify that the lock state is still consistent if the user resolves the `unifiedClasspath` task, either intentionally or through a plugin like [palantir/gradle-configuration-resolver-plugin](https://github.com/palantir/gradle-configuration-resolver-plugin), but the latter comes with the downside that it potentially downloads a lot of stuff you don't need, since it resolves all configurations.

On the other hand, resolving just `unifiedClasspath` is a bit annoying, you can't even run `dependencies --configuration unifiedClasspath` because we special cased that to not fail. So the alternative is to create a custom task that resolves unifiedClasspath - not great.

## After this PR
We now have a bespoke task for verifying that locks are still consistent, `verifyLocks`.

==COMMIT_MSG==
Lock verification can now be triggered via the `verifyLocks` task, which the lifecycle `check` task depends on.

WARNING: Locks are no longer verified for consistency as part of resolving unifiedClasspath
==COMMIT_MSG==

## Possible downsides?
* It might be confusing that lock verification now happens at a different point than before. However, it should be a lot more controllable now.

### DIFF
--- a/src/main/java/com/palantir/gradle/versions/VerifyLocksTask.java
+++ b/src/main/java/com/palantir/gradle/versions/VerifyLocksTask.java
@@ -1,0 +1,93 @@
+/*
+ * (c) Copyright 2019 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.gradle.versions;
+
+import com.google.common.collect.MapDifference;
+import com.google.common.collect.MapDifference.ValueDifference;
+import com.google.common.collect.Maps;
+import com.palantir.gradle.versions.internal.MyModuleIdentifier;
+import com.palantir.gradle.versions.lockstate.Line;
+import com.palantir.gradle.versions.lockstate.LockState;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import org.gradle.api.DefaultTask;
+import org.gradle.api.provider.Property;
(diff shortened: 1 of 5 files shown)


src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -398,7 +366,7 @@ private static boolean haveSameGroupAndName(Project project, Project subproject)
                 && project.getGroup().equals(subproject.getGroup());
     }
 
-    private void failIfAnyDependenciesUnresolved(ResolvableDependencies resolvableDependencies) {
+    final void failIfAnyDependenciesUnresolved(ResolvableDependencies resolvableDependencies) {
```
iamdanfox: Put back to private?

src/main/java/com/palantir/gradle/versions/VerifyLocksTask.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import org.gradle.api.DefaultTask;
+import org.gradle.api.provider.Property;
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.TaskAction;
+import org.gradle.language.base.plugins.LifecycleBasePlugin;
+
+public class VerifyLocksTask extends DefaultTask {
```
iamdanfox, line 38: Could we have a cheeky little test that runs this twice and verifies the second time it is UP-TO-DATE?

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -192,15 +188,24 @@ public final void apply(Project project) {
             ImmutableList<String> tasks = ImmutableList.of(":dependencyInsight", ":dependencies");
             unifiedClasspath.getIncoming().afterResolve(r -> {
                 if (tasks.stream().anyMatch(project.getGradle().getTaskGraph()::hasTask)) {
-                    log.lifecycle("Not checking validity of locks since we are running tasks that inspect "
+                    log.warn("Not checking validity of locks since we are running tasks that inspect "
                             + "dependencies");
                     return;
                 }
                 failIfAnyDependenciesUnresolved(r);
```
iamdanfox: Let's move this verification to the task too, so it all happens in one place

iamdanfox (approved):

Issue #117, opened 2019-05-14
"Found dependencies whose dependents changed" but ./gradlew --write-locks doesn't fix it


iamdanfox:
## What happened?

@mglazer reported in `#dev-foundry-infra` that CI for his project was failing with:

```
Found dependencies whose dependents changed: {org.eclipse.jgit:org.eclipse.jgit=(Line{group=org.eclipse.jgit, name=org.eclipse.jgit, version=5.3.1.201904271842-r, numDependents=1, dependentsHash=40088599}, Line{group=org.eclipse.jgit, name=org.eclipse.jgit, version=5.3.1.201904271842-r, numDependents=1, dependentsHash=dc0be413})}. Please run './gradlew --write-locks'.
```

But running `./gradlew --write-locks` didn't make any changes.

## What did you want to happen?

This kind of error message should never happen! Maybe there should be a clearer 'verify lock file' command instead of re-purposing the `resolveConfigurations` command?

dansanduleac:
Fixed by #119

PR #125, opened 2019-05-16
[fix] Fix locked versions feeding back into new lock state

dansanduleac:
## Before this PR

Addresses another root cause of https://github.com/palantir/gradle-consistent-versions/issues/117.
Specifically, when deciding the "current" lock state as part of the `verifyLocks` task, current constraints coming from `vesions.lock` are taken into account when doing that resolution.

As an example, a version of `1.2.0` from `versions.lock` will artifically raise the version in the "current lock state", even if the constraints have since changed such that they only require `1.1.0`. In that case, `verifyLocks` would think everything is up to date, even if `--write-locks` would actually produce a different lock file (with the correct new version of `1.1.0`).

## After this PR
==COMMIT_MSG==
Fix locked versions feeding back into new lock state. `verifyLocks` determines current lock state completely independently of the current versions in `versions.lock`.
==COMMIT_MSG==

We also simplified how we configure the recursive dependency copying for unifiedClasspath - instead of relying on the laziness of `withDependencies`, which was hard to follow, we just call the `recursivelyCopyProjectDependencies` method at the appropriate time in each case (writing locks vs not).

### DIFF
--- a/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
@@ -195,8 +195,12 @@ public class VersionsLockPlugin implements Plugin<Project> {
                 check.dependsOn(verifyLocks);
             });
 
-            // Can configure using constraints immediately, because rootLockfile exists.
-            configureAllProjectsUsingConstraints(project, rootLockfile);
+            project.getGradle().projectsEvaluated(g -> {
+                // Before we configure constraints, we must ensure that unifiedClasspath has scoured all projects.
+                unifiedClasspath.getIncoming().getDependencies();
+                // Can configure using constraints immediately, because rootLockfile exists.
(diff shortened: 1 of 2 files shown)


src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -298,7 +302,7 @@ private void recursivelyCopyProjectDependencies(
             Project currentProject, DependencySet dependencySet, Map<Configuration, String> copiedConfigurationsCache) {
         dependencySet
                 .matching(dependency -> ProjectDependency.class.isAssignableFrom(dependency.getClass()))
-                .configureEach(dependency -> {
+                .all(dependency -> {
```
iamdanfox, line 306: configureEach is lazy and all is not lazy, and apparently this change wasn't super essential but nice to be defensive anyway.

iamdanfox (approved):

dansanduleac:
👍

PR #134, opened 2019-05-30
[improvement] Make dependency sourcing more robust

dansanduleac:
# Before this PR
We collect the dependencies that we are interested in locking from each project via a `subprojectUnifiedClasspath` configuration.
However, the way we collect these is by making `subprojectUnifiedClasspath` just extend the other configurations we're interested in (`compileClasspath`, `runtimeClasspath`). 

This has a few problems:
* an exclude in one configuration propagates into `subprojectUnifiedClasspath`, potentially concealing a real dependency (that only appears in runtime) from the lock file
* since we copy and flatten `subprojectUnifiedClasspath`, we can't tell what configuration each dependency came from anymore. So for instance, if we were to also lock `testCompileClasspath` in this way, we couldn't trace if a dependency only came from test, or is also used in production.

Additionally, we find and copy every inter-project dependency declared by the user, and make an unsafe assumption that if the dependency didn't have a target configuration, then it must be the `default` configuration.

## After this PR
==COMMIT_MSG==
When picking up project dependencies, separate them into configurations based on their source (compile, runtime, + test etc in the future), such that gradle resolves them separately. This makes it less likely that additional rules (e.g. excludes) leak from one configuration into another and poison the final lock state for that project.
==COMMIT_MSG==

We also remove the need to copy the user's inter-project dependencies, by instead relying on [attribute-based matching](https://docs.gradle.org/current/userguide/dependency_management_attribute_based_matching.html) to control what configuration you get when encountering an inter-project dependency without a target configuration.

## Possible downsides?
* relying on more advanced gradle magic - [attribute-based matching](https://docs.gradle.org/current/userguide/dependency_management_attribute_based_matching.html), although this has been introduced for a while now and a core part of dependency resolution in java projects - this is how gradle knows whether to use the compile or runtime variant of a dependency & its transitives


src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -201,36 +248,67 @@ private static void sourceDependenciesFromProject(
...
-        // dependencies they have.
-        project.getPluginManager().withPlugin("base", plugin -> {
-            subprojectUnifiedClasspath.configure(conf -> {
-                conf.extendsFrom(project.getConfigurations().getByName(Dependency.DEFAULT_CONFIGURATION));
-            });
+        project.getConfigurations().register(SUBPROJECT_UNIFIED_CONFIGURATION_NAME, conf -> {
+            conf.setVisible(false).setCanBeResolved(false);
+
+            // Mark it so it doesn't receive constraints from VersionsPropsPlugin
+            conf.getAttributes().attribute(VersionsPropsPlugin.CONFIGURATION_EXCLUDE_ATTRIBUTE, true);
```
ferozco, line 254: I think just using configuration names to determine what configurations to exclude might be more straightforward/readable. Perhaps as a FLUP

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -201,36 +248,67 @@ private static void sourceDependenciesFromProject(
...
+
+            // Mark it as a GCV_SOURCE, so that
+            // 1. it can be depended on when resolving `unifiedClasspath`, which requests that the usage is GCV_SOURCE
```
ferozco: Is it true that the attribute needs to be set? I think that the second point is really why we want to set the attribute

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -201,36 +248,67 @@ private static void sourceDependenciesFromProject(
...
+                .add(createConfigurationDependency(project, SUBPROJECT_UNIFIED_CONFIGURATION_NAME));
+
         project.getPluginManager().withPlugin("java", plugin -> {
-            project.getConfigurations().named(SUBPROJECT_UNIFIED_CONFIGURATION_NAME).configure(conf -> {
-                Stream.of(
-                        JavaPlugin.COMPILE_CLASSPATH_CONFIGURATION_NAME,
-                        JavaPlugin.RUNTIME_CLASSPATH_CONFIGURATION_NAME)
-                        .map(project.getConfigurations()::getByName)
-                        .forEach(conf::extendsFrom);
+            String compileClasspathForLock = "consistentVersionsCompile";
```
ferozco: variableName == to string?

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -318,9 +419,25 @@ private void recursivelyCopyProjectDependencies(
...
+                    }
+
+                    log.info(
+                            "Found legacy project dependency (with target configuration): {} -> {}",
+                            dependencySet,
+                            formatProjectDependency(projectDependency));
+
                     if (copiedConfigurationsCache.containsKey(targetConf)) {
                         String copiedConf = copiedConfigurationsCache.get(targetConf);
                         log.debug("Re-using already copied target configuration for dep {} -> {}: {}",
```
ferozco, line 426: line 449: mark copied configuration as GCV_INTERNAL

ferozco: +1 for test too

src/test/groovy/com/palantir/gradle/versions/VersionsLockPluginIntegrationSpec.groovy
```diff
@@ -51,6 +51,7 @@ class VersionsLockPluginIntegrationSpec extends IntegrationSpec {
                 repositories {
                     maven { url "file:///${mavenRepo.getAbsolutePath()}" }
                 }
+                group = "test"
```
ferozco: do we need this?

src/test/groovy/com/palantir/gradle/versions/VersionsLockPluginIntegrationSpec.groovy
```diff
@@ -420,9 +421,10 @@ class VersionsLockPluginIntegrationSpec extends IntegrationSpec {
         new File(subdir, 'build.gradle') << """
             apply plugin: 'java-library'
         """.stripIndent()
+        debug = true
```
ferozco: nit: ensure to remove before merge

src/test/groovy/com/palantir/gradle/versions/VersionsLockPluginIntegrationSpec.groovy
```diff
@@ -420,9 +421,10 @@ class VersionsLockPluginIntegrationSpec extends IntegrationSpec {
...
 
         expect:
-        runTasks('--write-locks')
+        runTasks('--write-locks', '-S')
```
ferozco: samesies

ferozco (commented):
looking pretty hot!

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
dansanduleac: so turns out it's not a good idea to mark everything ORIGINAL as this then gets published in the module metadata file and causes potential issues for consumers in the future. Unless we only put that attribute onto other configurations that are not `runtimeElements` / `apiElements`, but then what's the point

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
dansanduleac: Will do this as a FLUP

iamdanfox:
10/10 PR desc mate.

iamdanfox (approved):

PR #141, opened 2019-06-04
[feature] lock test dependencies

### LINKED ISSUE #131: Tracking: lock test dependencies
## What happened?

Currently, we only only track production dependencies in `versions.lock`.
This is great for looking up at a glance what dependencies will be used in production, but it might hide the fact that tests are passing because one of the production dependencies' version got bumped transitively through some other testing utilities library.

Lock test dependencies too, but in a way that's clearly separated from the production dependencies.
As in, it should be obvious looking at the lock file which dependencies are used in production, and which are only used in tests.

Checklist:

- [ ] investigate how to track dependency provenance (e.g. it only came from `testCompileClasspath`)
- [ ] devise lock file structure changes to accommodate a separation of dependencies used in production / test / maybe other source sets

dansanduleac:
## Before this PR

We are only locking production (`compileClasspath`, `runtimeClasspath`) dependencies of every java project.
However, we are not locking test dependencies.
This means that a test dependency could accidentally upgrade the versions of jars used in production, and then the tests would run (with the `testRuntimeClasspath` classpath) using different versions than what is used in production.
This in turn means that the tests could be passing because of the upgraded versions, whereas if _exactly_ the versions as declared in the lock file were used, they might not pass.

## After this PR
==COMMIT_MSG==
Test dependencies are now locked together with production dependencies, but dependencies used _only_ in test are written in a separate block.
==COMMIT_MSG==

fixes https://github.com/palantir/gradle-consistent-versions/issues/131

## Possible downsides?
* users' lockfiles will most likely change


src/main/java/com/palantir/gradle/versions/VersionsLockExtension.java
```diff
@@ -0,0 +1,70 @@
...
+
+package com.palantir.gradle.versions;
+
+import java.util.stream.Stream;
+import javax.inject.Inject;
+import org.gradle.api.Project;
+import org.gradle.api.provider.SetProperty;
+import org.gradle.api.tasks.SourceSet;
+
+public class VersionsLockExtension {
```
iamdanfox, line 29: Something like?
```gradle
            versionsLock {
                disableDefaults()
                production {
                    configurations.eteTest
                    sourceSets.integrationTest
                }
                test {
                    
                }
            }
```

README.md
```diff
@@ -119,10 +121,16 @@ javax.inject:javax.inject:1 (2 constraints: d614a0ab)
...
+com.jayway.awaitility:awaitility:1.6.5 (1 constraints: c615c1d2)
 ```
 
-The lockfile is sourced from the _compileClasspath_ and _runtimeClasspath_ configurations.
-(Test-only dependencies will not appear in `versions.lock`)
+The lockfile sources production dependencies from the _compileClasspath_ and _runtimeClasspath_ configurations, and
+test dependencies from the compile/runtime classpaths of any source set that ends in test (e.g. `test`, `integrationTest`,
+`eteTest`). You can further [customize which source sets / configurations get locked](#common-workflow-locking-other-source-sets-or-configurations).
 
 <!-- TODO(dfox): build some ./gradlew checkVersionsLock task and recommend running it on CI -->
```
iamdanfox: this is the verifyLocks task???

README.md
```diff
@@ -256,6 +264,50 @@ Adding explicit `it` calls works around this error:
...
+`test` as test dependencies. This behaviour can be configured per project via the `versionsLock` extension.
+
+For instance, the default behaviour (assuming you only have the default `main` and `test` source sets) can be replicated through:
+
+```gradle
+versionsLock {
+    disableJavaPluginDefaults()  // do not lock any configuration / source set by default
+    production { from sourceSets.main }
+    test { from sourceSets.test }
+}
```
iamdanfox: I think all snippets should be copy-and-pasteable with some useful purpose - since this just replaces thedefaults, don't think it's worth including

README.md
```diff
@@ -256,6 +264,50 @@ Adding explicit `it` calls works around this error:
...
+
+```gradle
+# foo-test-common/build.gradle
+apply plugin: 'java'
+
+versionsLock {
+    disableJavaPluginDefaults()  // do not lock any configuration / source set by default
+    test {
+        from sourceSets.main
+    }
```
iamdanfox: I think we should actually make some little utility for people that says this entire project is a test project.

```gradle
versionsLock {
  testProject()
}
```

iamdanfox (approved):

dansanduleac:
👍

src/main/java/com/palantir/gradle/versions/GradleWorkarounds.java
```diff
@@ -51,6 +58,37 @@ static boolean isConfiguring(ProjectState state) {
         }
     }
 
+    /**
+     * Work around gradle < 5.3.rc-1 not adding an AttributeFactory to {@link ProjectDependency} with configuration,
+     * and {@link ExternalDependency#copy()} not configuring an AttributeFactory and ALSO not immutably copying the
+     * {@link AttributeContainer}.
+     */
+    static <T extends ModuleDependency> T fixAttributesOfModuleDependency(
```
dansanduleac, line 66: the reflective parts of this method can go away once this PR goes in: https://github.com/gradle/gradle/pull/9653

PR #160, opened 2019-06-26
Run tests using junit5

dansanduleac:
## Before this PR

Running tests using junit 4 runners.

## After this PR
==COMMIT_MSG==
Running tests on junit 5 vintage engine (still via spock runner).
==COMMIT_MSG==


build.gradle
```diff
@@ -10,34 +10,32 @@ buildscript {
...
-        classpath 'com.gradle.publish:plugin-publish-plugin:0.10.1'
         classpath 'com.netflix.nebula:nebula-publishing-plugin:12.0.0'
-        classpath 'com.palantir.baseline:gradle-baseline-java:0.62.1'
-        classpath 'com.palantir.gradle.consistentversions:gradle-consistent-versions:1.9.2'
-        classpath 'gradle.plugin.com.palantir.gradle.gitversion:gradle-git-version:0.7.3'
-        classpath 'gradle.plugin.org.inferred:gradle-processors:2.2.0'
     }
 }
 
+plugins {
```
ferozco, line 17: could we do this separately?

versions.props
```diff
@@ -2,6 +2,8 @@ com.netflix.nebula:nebula-dependency-recommender = 7.5.6
 com.netflix.nebula:nebula-test = 7.2.5
 com.google.guava:guava = 23.6.1-jre
 org.immutables:value = 2.7.5
+org.junit.jupiter:* = 5.4.0
```
ferozco: could you use the bom instead?

build.gradle
```diff
@@ -10,34 +10,32 @@ buildscript {
...
+    id 'com.jfrog.bintray' version '1.8.4' apply false
+    id 'com.gradle.plugin-publish' version '0.10.1'
+    id 'com.palantir.consistent-versions' version '1.9.2'
+    id 'org.inferred.processors' version '2.2.0'
+    id 'com.palantir.baseline' version '0.62.1'
+    id 'com.palantir.git-version' version '0.7.3'
+    id 'com.adarshr.test-logger' version '1.7.0'
```
iamdanfox: why com.adarshr.test-logger ?

versions.props
iamdanfox: Also 5.4.2? https://mvnrepository.com/artifact/org.junit.jupiter/junit-jupiter

build.gradle
dansanduleac: why not, provides pretty output for each test including how long it took, which I actually like

build.gradle
dansanduleac: don't think it's worth creating a new PR for this

build.gradle
```diff
@@ -46,12 +43,27 @@ dependencies {
...
+    useJUnitPlatform()
+
+    systemProperties = [
+            'junit.platform.output.capture.stdout'                  : 'true',
+            'junit.platform.output.capture.stderr'                  : 'true',
+            'junit.jupiter.execution.parallel.enabled'              : 'true',
+            'junit.jupiter.execution.parallel.config.strategy'      : 'dynamic',
+            'junit.jupiter.execution.parallel.config.dynamic.factor': '1'
+    ]
+}
```
iamdanfox, line 65: I think I'd like to bake this into baseline before we proliferate it any further.

build.gradle
```diff
@@ -46,12 +43,27 @@ dependencies {
...
+    testCompile platform('org.junit:junit-bom')
+    testImplementation 'com.netflix.nebula:nebula-test'
+    testImplementation "org.junit.jupiter:junit-jupiter-api"
+    testRuntimeOnly "org.junit.vintage:junit-vintage-engine"
 
     annotationProcessor "org.immutables:value"
     compileOnly "org.immutables:value::annotations"
 }
 
+tasks.withType(Test) {
```
robert3005, line 55: You likely want (or some version of)
```
testLogging {
    events "passed", "skipped", "failed"
}
```

as otherwise you get no summary when running (only what current gradle count displays)

iamdanfox: So I think I'm actually fine with that - you can look at the HTML report to see what's happened.

I want to keep CI output as high signal as possible, so that compiler warnings etc really stand out

build.gradle
robert3005: output capture is also pretty annoying if you want to have a quick look at what your test is doing. Likely most cases you don't care but if you do you have to wait until end of the run

iamdanfox: I've found filtering down using `--tests` pretty seamless for that?

Perhaps I could say locally do the testLogging but keep it clean on CI?

robert3005: Sure - you could then just log failed? Otherwise you won’t know until ci finishes

build.gradle
robert3005: Sure - you could then just log failed? Otherwise you won’t know until ci finishes

ferozco (approved):

build.gradle
dansanduleac: if tests take longer than 10 minutes, that's a bigger problem :)

PR #169, opened 2019-07-10
Mention need to use java-base with BOMs

dansanduleac:
## Before this PR

People are confused as to why they don't get constraints from their BOMs.

## After this PR
==COMMIT_MSG==
README now mentions the need to apply `java-base` on all projects in order for BOMs to work.
==COMMIT_MSG==


ferozco (approved):

ferozco:
👍

PR #187, opened 2019-08-14
More leninent checks

dansanduleac:
## Before this PR
Throwing if any configuration whatsoever uses `resolutionStrategy.failOnVersionConflict()`

## After this PR
==COMMIT_MSG==
Only throw on configurations using `resolutionStrategy.failOnVersionConflict()` if they are locked.

This allows configurations from random plugins / manually constructed by the user to continue existing as they are.
==COMMIT_MSG==

### DIFF
--- a/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
@@ -418,21 +418,6 @@ public class VersionsLockPlugin implements Plugin<Project> {
 
         project.subprojects(subproject -> {
             subproject.afterEvaluate(sub -> {
-                Configuration lockConstraints =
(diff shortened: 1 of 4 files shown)


src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -439,6 +430,16 @@ private static void checkPreconditions(Project project) {
         });
     }
 
+    private static void ensureNoFailOnVersionConflict(Configuration conf) {
+        org.gradle.api.internal.artifacts.configurations.ConflictResolution conflictResolution =
+                ((org.gradle.api.internal.artifacts.configurations.ResolutionStrategyInternal)
+                         conf.getResolutionStrategy()).getConflictResolution();
```
iamdanfox: maybe worth hiding this in the gradleworkarounds file?

iamdanfox (approved):

PR #188, opened 2019-08-14
Fail fast if multiple subprojects share same coordinate

dansanduleac:
## Before this PR
Only failing if a subproject shares the same `group:name` coordinate with the root project.

We didn't check if subprojects share a coordinate between themselves, which is also bad.

## After this PR
==COMMIT_MSG==
Fail if _any_ projects share the same group:name coordinate.
==COMMIT_MSG==

### DIFF
--- a/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
@@ -25,6 +25,9 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
+import com.google.common.collect.LinkedHashMultimap;
+import com.google.common.collect.Maps;
+import com.google.common.collect.Multimap;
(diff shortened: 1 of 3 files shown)


src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -406,13 +411,23 @@ private static void checkPreconditions(Project project) {
                 "Gradle Consistent Versions doesn't currently work with configure-on-demand, please remove"
                         + " 'org.gradle.configureondemand' from your gradle.properties");
 
+        Multimap<String, Project> coordinateDuplicates = LinkedHashMultimap.create();
+        Set<Project> subprojectsLeft = new HashSet<>(project.getSubprojects());
```
iamdanfox, line 420: is concurrency important here? surely subprojects will end up getting evaluated on different threads??

dansanduleac: I don't think so, that would break the world. Oddly I can't find written evidence of this though :(

ferozco:
👍

PR #191, opened 2019-08-15
Publish constraints filtered by transitive closure of dependencies

dansanduleac:
## Before this PR

We can't publish [gradle module metadata](https://blog.gradle.org/gradle-metadata-1.0) because:
* we'd publish constraints coming from `versions.props`, which is not ideal, because they might affect more components than just the "production" dependencies that we end up dependencing on from compile / runtime.
* we wouldn't publish constraints for transitive jars that your project actually depends on

## After this PR
==COMMIT_MSG==
Publish constraints from versions.lock filtered by the transitive closure of dependencies for each project.
==COMMIT_MSG==

A workaround has been introduced in GradleWorkarounds, but this PR should make the workaround unnecessary in future versions: https://github.com/gradle/gradle/pull/10288


src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -823,9 +885,30 @@ private static LockedConfigurations computeConfigurationsToLock(Project project,
...
     }
+
+    private static List<DependencyConstraint> constructPublishableConstraintsFromLockFile(
+            Path gradleLockfile, DependencyConstraintHandler constraintHandler) {
+        LockState lockState = new ConflictSafeLockFile(gradleLockfile).readLocks();
+        // We only publish the production locks.
+        return lockState.productionLinesByModuleIdentifier().entrySet().stream()
+                .map(e -> e.getKey() + ":" + e.getValue().version())
+                // Note: constraints.create sets the version as preferred + required, we want 'strictly' just like
+                // gradle does when verifying a lock file.
```
iamdanfox: delete comment?

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -823,9 +885,30 @@ private static LockedConfigurations computeConfigurationsToLock(Project project,
...
+                .map(notation -> constraintHandler.create(notation, constraint -> {
+                    constraint.version(v -> {
+                        String version = Objects.requireNonNull(constraint.getVersion());
+                        v.require(version);
+                    });
+                    constraint.because("Computed from versions.lock");
```
iamdanfox: mention gradle-consistent-versions?

iamdanfox (approved):

PR #193, opened 2019-08-16
Don't publish constraints from versions.props

dansanduleac:
## Before this PR

#191 failed to properly remove versions.props constraints from being published.

This is because they were still being inherited by the published `apiElements`, `runtimeElements` configurations, through other configurations like `runtime`, `implementation` etc.

## After this PR
==COMMIT_MSG==
Properly remove versions.props constraints from being published. This makes us now compatible with gradle metadata.

This is done by not configuring versions.props constraints on _any_ configuration that `apiElements` or `runtimeElements` extends from.

In order to maintain resolvability of the legacy configurations `compile`, `runtime` and `compileOnly`, we inject their versions at resolution time in the same way that nebula-dependency-recommender works.
We can't apply constraints on them anymore because they ancestors of the published configurations and so those constraints would be published.
==COMMIT_MSG==

### DIFF
--- a/src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
@@ -48,6 +48,9 @@ public class VersionsPropsPlugin implements Plugin<Project> {
     private static final Logger log = Logging.getLogger(VersionsPropsPlugin.class);
     private static final String ROOT_CONFIGURATION_NAME = "rootConfiguration";
     private static final GradleVersion MINIMUM_GRADLE_VERSION = GradleVersion.version("5.1");
+    private static final ImmutableList<String> JAVA_PUBLISHED_CONFIGURATION_NAMES = ImmutableList.of(
+            JavaPlugin.RUNTIME_ELEMENTS_CONFIGURATION_NAME,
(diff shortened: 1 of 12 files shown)


dansanduleac:
Remaining work necessary here:
ensure we inject versions into all published dependencies (platforms and otherwise). Should probably use gradle 5.3's versionMapping to ensure this.

src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
```diff
@@ -141,12 +121,28 @@ private static void setupConfiguration(
                 log.debug("Not configuring {} because it's excluded", conf);
                 return;
             }
-            // Because of https://github.com/gradle/gradle/issues/7954, we need to manually inject versions
-            // of direct dependencies if they come from a *-constraint
-            // Note: this is necessary on the rootConfiguration too in order to support injecting versions of
-            // BOMs.
+
+            if (JAVA_PUBLISHED_CONFIGURATION_NAMES.stream().anyMatch(conf.getName()::equals)) {
```
iamdanfox: set contains instead?

iamdanfox (approved):
LGTM, unlabel when you're ready

dansanduleac:
👍

PR #197, opened 2019-08-20
Fix other regression with BOMs

dansanduleac:
## Before this PR
#191 introduced a regression that broke injecting versions into BOMs declared in `rootConfiguration`.
#195 fixed this somewhat, but only for the locked configurations, thanks to the VersionsLockPlugin. Non-locked configurations like `conjureJava` still got the platform dependency and failed to resolve it (as evidenced by the [failing test](https://circleci.com/gh/palantir/gradle-consistent-versions/1876?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) in the first commit).

## After this PR
==COMMIT_MSG==
Fix regression from 1.11.0 where non-locked configurations could not resolve a `platform()` dependency if its constraint was declared using a `*` in `versions.props`.

Example:
```
Could not resolve all files for configuration ':some-api:conjureJava'.
   > Could not find com.palantir.witchcraft:witchcraft-core-bom:.
```
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-197.v2.yml
+++ b/changelog/@unreleased/pr-197.v2.yml
@@ -1,0 +1,12 @@
+type: fix
+fix:
+  description: |-
+    Fix regression from 1.11.0 where non-locked configurations could not resolve a `platform()` dependency if its constraint was declared using a `*` in `versions.props`.
+
+    Example:
+    ```
+    Could not resolve all files for configuration ':some-api:conjureJava'.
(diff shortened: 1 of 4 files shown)


iamdanfox (approved):
Sounds reasonable, but can we do a `publishToMavenLocal` and verify this fully fixes the regression on @mglazer's project that reported this?

PR #226, opened 2019-09-09
Move gradle plugins to buildscript

dansanduleac:
## Before this PR
Plugins defined in plugins DSL make it hard to automatically track adoption of plugin jars as the connection between a plugin and the jar where it is defined is indirect.

## After this PR
==COMMIT_MSG==
Move gradle plugins to buildscript so the actual jars used can be tracked
==COMMIT_MSG==

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,7 +15,7 @@ buildscript {
         classpath 'com.palantir.gradle.consistentversions:gradle-consistent-versions:1.12.1'
         classpath 'com.palantir.baseline:gradle-baseline-java:2.1.1'
         classpath 'com.palantir.gradle.gitversion:gradle-git-version:0.12.2'
-        classpath 'org.inferred:processors:3.1.0'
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -10,18 +10,18 @@ buildscript {
...
+        classpath 'com.netflix.nebula:gradle-info-plugin:5.0.3'
+        classpath 'com.palantir.gradle.consistentversions:gradle-consistent-versions:1.12.1'
+        classpath 'com.palantir.baseline:gradle-baseline-java:2.1.1'
+        classpath 'com.palantir.gradle.gitversion:gradle-git-version:0.12.2'
+        classpath 'gradle.plugin.org.inferred:gradle-processors:3.1.0'
     }
 }
 
 plugins {
     id 'com.jfrog.bintray' version '1.8.4' apply false
```
ferozco: can we move this one?

ferozco (approved):

PR #237, opened 2019-09-11
Fix: Don't fail if excluded configuration is resolved early

dansanduleac:
## Before this PR

There is a regression in 1.11 which occurs when using the `nebula.dependency-recommender` plugin.
It will attempt to resolve the `nebulaRecommenderBom` configuration early, which will then fail.

```
Caused by: org.gradle.api.GradleException: Not allowed to resolve configuration ':nebulaRecommenderBom' at configuration time (https://guides.gradle.org/performance/#don_t_resolve_dependencies_at_configuration_time). Please upgrade your plugins and double-check your gradle scripts (see stacktrace)
        at com.palantir.gradle.versions.VersionsPropsPlugin.lambda$setupConfiguration$7(VersionsPropsPlugin.java:161)
```

## After this PR
==COMMIT_MSG==
Fix regression where we were failing if an excluded configuration is resolved early.
==COMMIT_MSG==

### DIFF
--- a/src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
@@ -16,11 +16,13 @@
 
 package com.palantir.gradle.versions;
 
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
-
 import org.gradle.api.GradleException;
 import org.gradle.api.NamedDomainObjectProvider;
 import org.gradle.api.Plugin;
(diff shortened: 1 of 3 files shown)


src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
```diff
@@ -147,21 +147,21 @@ private static void setupConfiguration(
...
+
+            // Add fail-safe error reporting
+            conf.getIncoming().beforeResolve(resolvableDependencies -> {
+                if (GradleWorkarounds.isConfiguring(subproject.getState())) {
+                    throw new GradleException(String.format("Not allowed to resolve %s at "
+                            + "configuration time (https://guides.gradle.org/performance/"
+                            + "#don_t_resolve_dependencies_at_configuration_time). Please upgrade your "
+                            + "plugins and double-check your gradle scripts (see stacktrace)", conf));
+                }
+            });
```
dansanduleac, line 164: This move is safe.
This whole parent block (`withDependencies`) is guaranteed to be executed (by `runDependencyActions` in the linked code below) before the `beforeResolve` action is run.

https://github.com/gradle/gradle/blob/ed58dbc5366ddc542a655810235f68e83cf82584/subprojects/dependency-management/src/main/java/org/gradle/api/internal/artifacts/configurations/DefaultConfiguration.java#L620-L624

iamdanfox:
Given that this was already a regression, can we add a little test to make sure it doesnt' come back?

iamdanfox (approved):
Worth doing a pTML to see if this breaks anything unexpected?

PR #275, opened 2019-09-21
Use new version of GCV in README

zachkirsch:
## Before this PR
An old version of GCV is used in the README installation instructions. People that follow those steps to install will end up with a pretty old version of GCV.

## After this PR
==COMMIT_MSG==
README uses current version of GCV in the installation instructions.
==COMMIT_MSG==

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,10 @@ Direct dependencies are specified in a top level `versions.props` file and then 
 1. Apply the plugin (root project only):
     ```gradle
     plugins {
-        id "com.palantir.consistent-versions" version "1.12.4"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,7 +12,7 @@ Direct dependencies are specified in a top level `versions.props` file and then
 1. Apply the plugin (root project only):
     ```gradle
     plugins {
-        id "com.palantir.consistent-versions" version "1.4.0"
+        id "com.palantir.consistent-versions" version "1.12.4"
```
ferozco: This is going to get stale right away, instead of listing an exact version could we say `<current version>`

ferozco (approved):

PR #283, opened 2019-09-27
Exemplifying palantir-java-format changes

dansanduleac:
## After this PR
==COMMIT_MSG==

==COMMIT_MSG==

### DIFF
--- a/src/main/java/com/palantir/gradle/versions/ConflictSafeLockFile.java
+++ b/src/main/java/com/palantir/gradle/versions/ConflictSafeLockFile.java
@@ -47,9 +47,7 @@ final class ConflictSafeLockFile {
         this.lockfile = lockfile;
     }
 
-    /**
-     * Reads and returns the {@link LockState}.
-     */
+    /** Reads and returns the {@link LockState}. */
     public LockState readLocks() {
         try (Stream<String> linesStream = Files.lines(lockfile)) {
             List<String> lines = linesStream.filter(line -> !line.trim().startsWith("#")).collect(Collectors.toList());
@@ -57,8 +55,10 @@ final class ConflictSafeLockFile {
             Stream<String> productionDeps;
             Stream<String> testDeps;
             if (testDependenciesPosition >= 0) {
-                productionDeps = lines.subList(0, testDependenciesPosition - 1) // skip blank line before marker
-                        .stream();
+                productionDeps =
+                        lines
+                                .subList(0, testDependenciesPosition - 1) // skip blank line before marker
+                                .stream();
                 testDeps = lines.subList(testDependenciesPosition + 1, lines.size()).stream();
             } else {
                 productionDeps = lines.stream().filter(line -> !line.trim().startsWith("#"));
@@ -68,8 +68,7 @@ final class ConflictSafeLockFile {
             return LockState.from(parseLines(productionDeps), parseLines(testDeps));
         } catch (IOException e) {
             throw new GradleException(
-                    String.format("Couldn't load versions from palantir dependency lock file: %s", lockfile),
-                    e);
+                    String.format("Couldn't load versions from palantir dependency lock file: %s", lockfile), e);
         }
     }
 
@@ -94,11 +93,8 @@ final class ConflictSafeLockFile {
 
     public void writeLocks(FullLockState fullLockState) {
         LockState lockState = LockStates.toLockState(fullLockState);
-        try (
-                BufferedWriter writer = Files.newBufferedWriter(
-                        lockfile,
-                        StandardOpenOption.CREATE,
-                        StandardOpenOption.TRUNCATE_EXISTING)) {
+        try (BufferedWriter writer =
+                Files.newBufferedWriter(lockfile, StandardOpenOption.CREATE, StandardOpenOption.TRUNCATE_EXISTING)) {
             writer.append(HEADER_COMMENT);
             writer.newLine();
 
--- a/src/main/java/com/palantir/gradle/versions/ConsistentVersionsPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/ConsistentVersionsPlugin.java
@@ -31,10 +31,9 @@ public class ConsistentVersionsPlugin implements Plugin<Project> {
         project.getPluginManager().apply(GetVersionPlugin.class);
 
         project.allprojects(proj -> {
-            proj.getPluginManager()
-                    .withPlugin("java", plugin -> {
-                        proj.getPluginManager().apply(FixLegacyJavaConfigurationsPlugin.class);
-                    });
+            proj.getPluginManager().withPlugin("java", plugin -> {
+                proj.getPluginManager().apply(FixLegacyJavaConfigurationsPlugin.class);
+            });
         });
     }
 }
--- a/src/main/java/com/palantir/gradle/versions/ConsistentVersionsPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/ConsistentVersionsPlugin.java
@@ -31,10 +31,10 @@ public class ConsistentVersionsPlugin implements Plugin<Project> {
         project.getPluginManager().apply(GetVersionPlugin.class);
 
         project.allprojects(proj -> {
-            proj.getPluginManager().withPlugin("java", plugin -> {
-                proj.getPluginManager().apply(FixLegacyJavaConfigurationsPlugin.class);
-            });
+            proj.getPluginManager()
+                    .withPlugin("java", plugin -> {
+                        proj.getPluginManager().apply(FixLegacyJavaConfigurationsPlugin.class);
+                    });
         });
     }
-
 }
--- a/src/main/java/com/palantir/gradle/versions/FixLegacyJavaConfigurationsPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/FixLegacyJavaConfigurationsPlugin.java
@@ -28,9 +28,9 @@ import org.gradle.api.artifacts.ResolvableDependencies;
 import org.gradle.api.plugins.JavaPlugin;
 
 /**
- * This plugin exists in order to ensure versions in legacy java configurations (that are still
- * {@link Configuration#isCanBeResolved() resolvable} for compatibility reasons) can be resolved, and that they use
- * the locked versions.
+ * This plugin exists in order to ensure versions in legacy java configurations (that are still {@link
+ * Configuration#isCanBeResolved() resolvable} for compatibility reasons) can be resolved, and that they use the locked
+ * versions.
  */
 public class FixLegacyJavaConfigurationsPlugin implements Plugin<Project> {
     @Override
@@ -44,31 +44,22 @@ public class FixLegacyJavaConfigurationsPlugin implements Plugin<Project> {
             return;
         }
 
-        Configuration unifiedClasspath = project
-                .getRootProject()
+        Configuration unifiedClasspath = project.getRootProject()
                 .getConfigurations()
                 .findByName(VersionsLockPlugin.UNIFIED_CLASSPATH_CONFIGURATION_NAME);
         Preconditions.checkNotNull(
-                unifiedClasspath,
-                "FixLegacyJavaConfigurationsPlugin must be applied after VersionsLockPlugin");
+                unifiedClasspath, "FixLegacyJavaConfigurationsPlugin must be applied after VersionsLockPlugin");
 
         fixLegacyResolvableJavaConfigurations(project, unifiedClasspath);
     }
 
     private void fixLegacyResolvableJavaConfigurations(Project project, Configuration unifiedClasspath) {
-        Stream.of(
-                JavaPlugin.COMPILE_CONFIGURATION_NAME,
-                JavaPlugin.COMPILE_ONLY_CONFIGURATION_NAME,
-                JavaPlugin.RUNTIME_CONFIGURATION_NAME)
+        Stream.of(JavaPlugin.COMPILE_CONFIGURATION_NAME, JavaPlugin.COMPILE_ONLY_CONFIGURATION_NAME, JavaPlugin
+                .RUNTIME_CONFIGURATION_NAME)
                 .map(project.getConfigurations()::named)
                 .forEach(confProvider -> confProvider.configure(conf -> {
-                    injectVersions(
-                            conf,
-                            (group, name) -> GetVersionPlugin.getOptionalVersion(
-                                    project,
-                                    group,
-                                    name,
-                                    unifiedClasspath));
+                    injectVersions(conf, (group, name) -> GetVersionPlugin.getOptionalVersion(
+                            project, group, name, unifiedClasspath));
                 }));
     }
 
@@ -76,9 +67,7 @@ public class FixLegacyJavaConfigurationsPlugin implements Plugin<Project> {
         Optional<String> getVersion(String group, String name);
     }
 
-    /**
-     * Inject versions of _all_ dependencies into the given {@code conf}, by polling the {@code getVersion}.
-     */
+    /** Inject versions of _all_ dependencies into the given {@code conf}, by polling the {@code getVersion}. */
     private void injectVersions(Configuration conf, GetVersion getVersion) {
         ResolvableDependencies incoming = conf.getIncoming();
         incoming.beforeResolve(dependencies -> {
@@ -88,26 +77,28 @@ public class FixLegacyJavaConfigurationsPlugin implements Plugin<Project> {
             }
             // Code adapted from:
             // https://github.com/nebula-plugins/nebula-dependency-recommender-plugin/blob/64ed7c6853f80b909918e6a595231a5e9803ae8b/src/main/groovy/netflix/nebula/dependency/recommender/DependencyRecommendationsPlugin.java
-            conf.getResolutionStrategy().eachDependency(details -> {
-                ModuleVersionSelector requested = details.getTarget();
+            conf.getResolutionStrategy()
+                    .eachDependency(details -> {
+                        ModuleVersionSelector requested = details.getTarget();
 
-                // don't interfere with the way forces trump everything
-                for (ModuleVersionSelector force : conf.getResolutionStrategy().getForcedModules()) {
-                    if (requested.getGroup().equals(force.getGroup()) && requested.getName().equals(force.getName())) {
-                        details.because(String.format(
-                                "Would have recommended a version for %s:%s, but a force is in place",
-                                requested.getGroup(),
-                                requested.getName()));
-                        return;
-                    }
-                }
+                        // don't interfere with the way forces trump everything
+                        for (ModuleVersionSelector force : conf.getResolutionStrategy().getForcedModules()) {
+                            if (requested.getGroup().equals(force.getGroup())
+                                    && requested.getName().equals(force.getName())) {
+                                details.because(String.format(
+                                        "Would have recommended a version for %s:%s, but a force is in place",
+                                        requested.getGroup(), requested.getName()));
+                                return;
+                            }
+                        }
 
-                getVersion.getVersion(details.getRequested().getGroup(), details.getRequested().getName())
-                        .ifPresent(ver -> {
-                            details.useVersion(ver);
-                            details.because("Forced by gradle-consistent-versions versions.lock");
-                        });
-            });
+                        getVersion
+                                .getVersion(details.getRequested().getGroup(), details.getRequested().getName())
+                                .ifPresent(ver -> {
+                                    details.useVersion(ver);
+                                    details.because("Forced by gradle-consistent-versions versions.lock");
+                                });
+                    });
         });
     }
 }
--- a/src/main/java/com/palantir/gradle/versions/FixLegacyJavaConfigurationsPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/FixLegacyJavaConfigurationsPlugin.java
@@ -44,9 +44,8 @@ public class FixLegacyJavaConfigurationsPlugin implements Plugin<Project> {
             return;
         }
 
-        Configuration unifiedClasspath = project.getRootProject()
-                .getConfigurations()
-                .findByName(VersionsLockPlugin.UNIFIED_CLASSPATH_CONFIGURATION_NAME);
+        Configuration unifiedClasspath = project.getRootProject().getConfigurations().findByName(VersionsLockPlugin
+                .UNIFIED_CLASSPATH_CONFIGURATION_NAME);
         Preconditions.checkNotNull(
                 unifiedClasspath, "FixLegacyJavaConfigurationsPlugin must be applied after VersionsLockPlugin");
 
@@ -77,28 +76,25 @@ public class FixLegacyJavaConfigurationsPlugin implements Plugin<Project> {
             }
             // Code adapted from:
             // https://github.com/nebula-plugins/nebula-dependency-recommender-plugin/blob/64ed7c6853f80b909918e6a595231a5e9803ae8b/src/main/groovy/netflix/nebula/dependency/recommender/DependencyRecommendationsPlugin.java
-            conf.getResolutionStrategy()
-                    .eachDependency(details -> {
-                        ModuleVersionSelector requested = details.getTarget();
+            conf.getResolutionStrategy().eachDependency(details -> {
+                ModuleVersionSelector requested = details.getTarget();
 
-                        // don't interfere with the way forces trump everything
-                        for (ModuleVersionSelector force : conf.getResolutionStrategy().getForcedModules()) {
-                            if (requested.getGroup().equals(force.getGroup())
-                                    && requested.getName().equals(force.getName())) {
-                                details.because(String.format(
-                                        "Would have recommended a version for %s:%s, but a force is in place",
-                                        requested.getGroup(), requested.getName()));
-                                return;
-                            }
-                        }
+                // don't interfere with the way forces trump everything
+                for (ModuleVersionSelector force : conf.getResolutionStrategy().getForcedModules()) {
+                    if (requested.getGroup().equals(force.getGroup()) && requested.getName().equals(force.getName())) {
+                        details.because(String.format(
+                                "Would have recommended a version for %s:%s, but a force is in place",
+                                requested.getGroup(), requested.getName()));
+                        return;
+                    }
+                }
 
-                        getVersion
-                                .getVersion(details.getRequested().getGroup(), details.getRequested().getName())
-                                .ifPresent(ver -> {
-                                    details.useVersion(ver);
-                                    details.because("Forced by gradle-consistent-versions versions.lock");
-                                });
-                    });
+                getVersion.getVersion(details.getRequested().getGroup(), details.getRequested().getName()).ifPresent(
+                        ver -> {
+                            details.useVersion(ver);
+                            details.because("Forced by gradle-consistent-versions versions.lock");
+                        });
+            });
         });
     }
 }
--- a/src/main/java/com/palantir/gradle/versions/FixLegacyJavaConfigurationsPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/FixLegacyJavaConfigurationsPlugin.java
@@ -44,8 +44,8 @@ public class FixLegacyJavaConfigurationsPlugin implements Plugin<Project> {
             return;
         }
 
-        Configuration unifiedClasspath = project.getRootProject().getConfigurations().findByName(VersionsLockPlugin
-                .UNIFIED_CLASSPATH_CONFIGURATION_NAME);
+        Configuration unifiedClasspath = project.getRootProject().getConfigurations().findByName(
+                VersionsLockPlugin.UNIFIED_CLASSPATH_CONFIGURATION_NAME);
         Preconditions.checkNotNull(
                 unifiedClasspath, "FixLegacyJavaConfigurationsPlugin must be applied after VersionsLockPlugin");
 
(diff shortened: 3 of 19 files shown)


src/main/java/com/palantir/gradle/versions/ConsistentVersionsPlugin.java
```diff
@@ -31,10 +31,10 @@ public final void apply(Project project) {
         project.getPluginManager().apply(GetVersionPlugin.class);
 
         project.allprojects(proj -> {
-            proj.getPluginManager().withPlugin("java", plugin -> {
-                proj.getPluginManager().apply(FixLegacyJavaConfigurationsPlugin.class);
-            });
+            proj.getPluginManager()
+                    .withPlugin("java", plugin -> {
```
dansanduleac: Laying out this kind of chained method calls on a single line doesn't work yet but is fixable.

src/main/java/com/palantir/gradle/versions/FixLegacyJavaConfigurationsPlugin.java
```diff
@@ -28,9 +28,9 @@
 import org.gradle.api.plugins.JavaPlugin;
 
 /**
- * This plugin exists in order to ensure versions in legacy java configurations (that are still
- * {@link Configuration#isCanBeResolved() resolvable} for compatibility reasons) can be resolved, and that they use
- * the locked versions.
+ * This plugin exists in order to ensure versions in legacy java configurations (that are still {@link
```
carterkozak, line 31: I find something deeply unsettling about line splits in javadoc tags. It's totally irrational, and not worth thinking about...

src/main/java/com/palantir/gradle/versions/FixLegacyJavaConfigurationsPlugin.java
```diff
@@ -44,41 +44,30 @@ public final void apply(Project project) {
...
         fixLegacyResolvableJavaConfigurations(project, unifiedClasspath);
     }
 
     private void fixLegacyResolvableJavaConfigurations(Project project, Configuration unifiedClasspath) {
-        Stream.of(
-                JavaPlugin.COMPILE_CONFIGURATION_NAME,
-                JavaPlugin.COMPILE_ONLY_CONFIGURATION_NAME,
-                JavaPlugin.RUNTIME_CONFIGURATION_NAME)
+        Stream.of(JavaPlugin.COMPILE_CONFIGURATION_NAME, JavaPlugin.COMPILE_ONLY_CONFIGURATION_NAME, JavaPlugin
+                .RUNTIME_CONFIGURATION_NAME)
```
dansanduleac: Currently it will try to lay out all arguments on the 2nd line if possible, but if not, it will prefer breaking the last argument in certain cases (if it's a lambda, method args or generally anything that includes a dot).
This can probably be made stricter so that field accesses like this one aren't allowed to be broken this way.

src/main/java/com/palantir/gradle/versions/FixLegacyJavaConfigurationsPlugin.java
```diff
@@ -44,41 +44,30 @@ public final void apply(Project project) {
...
 
-        Configuration unifiedClasspath = project
-                .getRootProject()
+        Configuration unifiedClasspath = project.getRootProject()
                 .getConfigurations()
                 .findByName(VersionsLockPlugin.UNIFIED_CLASSPATH_CONFIGURATION_NAME);
         Preconditions.checkNotNull(
-                unifiedClasspath,
-                "FixLegacyJavaConfigurationsPlugin must be applied after VersionsLockPlugin");
+                unifiedClasspath, "FixLegacyJavaConfigurationsPlugin must be applied after VersionsLockPlugin");
```
iamdanfox, line 50: :+1: this is nice - laying out method invocation args on one line where possible rather than two!

src/main/java/com/palantir/gradle/versions/GetVersionPlugin.java
```diff
@@ -81,18 +82,17 @@ private static String getVersion(Project project, String group, String name, Con
     }
 
     static Optional<String> getOptionalVersion(
-            Project project,
-            String group,
-            String name,
-            Configuration configuration) {
+            Project project, String group, String name, Configuration configuration) {
```
iamdanfox, line 76: 🤩

src/main/java/com/palantir/gradle/versions/GetVersionPlugin.java
```diff
@@ -81,18 +82,17 @@ private static String getVersion(Project project, String group, String name, Con
...
         if (GradleWorkarounds.isConfiguring(project.getState())) {
-            throw new GradleException(
-                    String.format("Not allowed to call gradle-consistent-versions's getVersion(\"%s\", \"%s\", "
+            throw new GradleException(String.format(
+                    "Not allowed to call gradle-consistent-versions's getVersion(\"%s\", \"%s\", "
                             + "configurations.%s) "
-                            + "at configuration time", group, name, configuration.getName()));
+                            + "at configuration time",
+                    group, name, configuration.getName()));
```
iamdanfox, line 82: This is nice - I think humans do this heuristic too

src/main/java/com/palantir/gradle/versions/GradleComparators.java
```diff
@@ -27,20 +27,17 @@
...
      */
-    public static final Comparator<ComponentIdentifier> COMPONENT_IDENTIFIER_COMPARATOR = Comparator
-            .comparing(
-                    (ComponentIdentifier id) -> tryCast(ModuleComponentIdentifier.class, id),
-                    Comparators.emptiesFirst(Ordering
-                            .from(MODULE_IDENTIFIER_COMPARATOR)
-                            .onResultOf(ModuleComponentIdentifier::getModuleIdentifier)))
+    public static final Comparator<ComponentIdentifier> COMPONENT_IDENTIFIER_COMPARATOR = Comparator.comparing(
+                    (ComponentIdentifier id) -> tryCast(ModuleComponentIdentifier.class, id), Comparators.emptiesFirst(
+                                    Ordering.from(MODULE_IDENTIFIER_COMPARATOR)
```
iamdanfox: I don't think this would pass code-review if a human produced it - maybe we'd need to extract out to a variable so that the formatter will leave it in a good state?

(easier reading here: https://github.com/palantir/gradle-consistent-versions/blob/db658f7babf946e5f85578586756311448821ce3/src/main/java/com/palantir/gradle/versions/GradleComparators.java)

src/main/java/com/palantir/gradle/versions/GradleWorkarounds.java
```diff
@@ -134,9 +132,7 @@ static boolean isConfiguring(ProjectState state) {
 
         try {
             Method method = org.gradle.api.internal.artifacts.dependencies.AbstractModuleDependency.class
-                    .getDeclaredMethod(
-                            "setAttributes",
-                            org.gradle.api.internal.attributes.AttributeContainerInternal.class);
+                    .getDeclaredMethod("setAttributes", org.gradle.api.internal.attributes.AttributeContainerInternal.class);
```
iamdanfox: I like that it collapses things onto one line where possible!

src/main/java/com/palantir/gradle/versions/GradleWorkarounds.java
```diff
@@ -71,52 +71,50 @@ static boolean isConfiguring(ProjectState state) {
...
-                                "Could not proxy method '%s' to object %s",
-                                method,
-                                property));
-                    } else {
-                        return method.invoke(property, args);
-                    }
-                });
+        return (ListProperty<T>)
+                Proxy.newProxyInstance(
+                        GradleWorkarounds.class.getClassLoader(),
```
iamdanfox, line 82: Kinda weird to break after the cast? We could save one layer of indenting if we could make the formatter pick this:

```java
        return (ListProperty<T>) Proxy.newProxyInstance(
                    GradleWorkarounds.class.getClassLoader(),
```

src/main/java/com/palantir/gradle/versions/GradleWorkarounds.java
```diff
@@ -169,9 +165,10 @@ private static boolean isPlatformPre53(AttributeContainer attributes) {
     }
 
     static boolean isFailOnVersionConflict(Configuration conf) {
-        org.gradle.api.internal.artifacts.configurations.ConflictResolution conflictResolution =
-                ((org.gradle.api.internal.artifacts.configurations.ResolutionStrategyInternal) conf
-                        .getResolutionStrategy()).getConflictResolution();
+        org.gradle.api.internal.artifacts.configurations.ConflictResolution conflictResolution = ((org.gradle.api
+                                .internal.artifacts.configurations.ResolutionStrategyInternal)
```
iamdanfox, line 170: wooooah not sure about this one.  breaking midway through a fully-qualified name seems not nice

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -278,138 +275,137 @@ public final void apply(Project project) {
...
-            LockedConfigurations lockedConfigurations = computeConfigurationsToLock(subproject, ext);
-            addConfigurationDependencies(subproject,
-                    subproject.getConfigurations().getByName(CONSISTENT_VERSIONS_PRODUCTION),
-                    lockedConfigurations.productionConfigurations());
-            addConfigurationDependencies(subproject,
-                    subproject.getConfigurations().getByName(CONSISTENT_VERSIONS_TEST),
-                    lockedConfigurations.testConfigurations());
-            return lockedConfigurations;
-        }));
+        return rootProject.getAllprojects().stream()
```
iamdanfox: does it have special-casing to put `.stream()` at the end of a contiguous line???

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -278,138 +275,137 @@ public final void apply(Project project) {
...
+                                .named("resolveConfigurations", task -> task.mustRunAfter(":resolveConfigurations"));
+                    });
         }
 
         // This is not how we collect dependencies, but is only meant to capture and neutralize the user's
         // inter-project dependencies.
-        project.getConfigurations().register(PLACEHOLDER_CONFIGURATION_NAME, conf -> {
-            conf.setVisible(false).setCanBeResolved(false);
+        project.getConfigurations()
+                .register(PLACEHOLDER_CONFIGURATION_NAME, conf -> {
```
iamdanfox: Interesting that a lot of the diff in this PR is because we seem to frequently have the pattern:

```java
foo.something().other().method("arg1", arg2, lambda -> {
  // lots of code
}
```

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -278,138 +275,137 @@ public final void apply(Project project) {
...
-                                project,
-                                consistentVersionsProduction.get(),
-                                GcvScope.PRODUCTION));
-        unifiedClasspath.getDependencies()
-                .add(
-                        createConfigurationDependencyWithScope(project, consistentVersionsTest.get(), GcvScope.TEST));
+        unifiedClasspath
+                .getDependencies()
+                .add(createConfigurationDependencyWithScope(project, consistentVersionsProduction.get(), GcvScope
+                        .PRODUCTION));
```
iamdanfox: On the whole I think these 4 lines are nicer, but breaking in the middle of the Enum.VARIANT is a bit weird

src/main/java/com/palantir/gradle/versions/VerifyLocksTask.java
```diff
@@ -74,45 +74,45 @@ public final void taskAction() throws IOException {
...
-                        diff.getValue().rightValue().stringRepresentation()))
+    private static String formatDependencyDifferences(Map<MyModuleIdentifier, ValueDifference<Line>> differing) {
+        return differing.entrySet().stream()
+                .map(diff -> String.format(
+                        "" // to align strings
+                                + "-%s\n"
+                                + "+%s",
+                        diff.getValue().leftValue().stringRepresentation(), diff.getValue()
+                                .rightValue()
+                                .stringRepresentation()))
```
dansanduleac: This is not great, it took a bunch of format arguments and split only the last one.
This behaviour should only be triggered if all the previous arguments fit on one line.
Will track this.

src/main/java/com/palantir/gradle/versions/ConflictSafeLockFile.java
```diff
@@ -47,18 +47,18 @@
...
     public LockState readLocks() {
         try (Stream<String> linesStream = Files.lines(lockfile)) {
             List<String> lines = linesStream.filter(line -> !line.trim().startsWith("#")).collect(Collectors.toList());
             int testDependenciesPosition = lines.indexOf(TEST_DEPENDENCIES_MARKER);
             Stream<String> productionDeps;
             Stream<String> testDeps;
             if (testDependenciesPosition >= 0) {
-                productionDeps = lines.subList(0, testDependenciesPosition - 1) // skip blank line before marker
-                        .stream();
+                productionDeps =
```
ferozco, line 58: sad days :(

src/main/java/com/palantir/gradle/versions/VersionsProps.java
```diff
@@ -56,9 +53,9 @@ public static VersionsProps loadFromFile(Path path) {
         FuzzyPatternResolver.Builder builder = FuzzyPatternResolver.builder();
         recommendations
                 .stringPropertyNames()
-                .forEach(name -> builder.putVersions(
-                        name.replaceAll("/", ":"),
-                        recommendations.getProperty(name).trim()));
+                .forEach(name -> builder.putVersions(name.replaceAll("/", ":"), recommendations
+                        .getProperty(name)
+                        .trim()));
```
dansanduleac: This is worse, but I think it might require an exponential approach to fix: try both splitting the last argument and every argument on its own line, choose the one that's better on some metric like LOC or max indent

src/main/java/com/palantir/gradle/versions/FixLegacyJavaConfigurationsPlugin.java
```diff
@@ -44,41 +44,30 @@ public final void apply(Project project) {
...
                 .forEach(confProvider -> confProvider.configure(conf -> {
-                    injectVersions(
-                            conf,
-                            (group, name) -> GetVersionPlugin.getOptionalVersion(
-                                    project,
-                                    group,
-                                    name,
-                                    unifiedClasspath));
+                    injectVersions(conf, (group, name) -> GetVersionPlugin.getOptionalVersion(
+                            project, group, name, unifiedClasspath));
```
ferozco, line 63: 👍

iamdanfox:
Overall I think the really big wins here are:

- denser method declaration and method invocation (aka jared style). You can see this diff is +545 −637
- it actually merges unbreaks lines where it thinks it makes sense (meaning we can fix up cases the eclipse-formatter kinda borked)
- if we have a fork of google-java-format, then we're empowered to make improvements rather than just helplessly twiddle knobs of eclipse-java-formatter
- this approach is zero config, so we actually would pre-empty some bikeshedding

downsides

- we're gonna have to publish IDE plugins for both intellij and eclipse
- this would mean actually maintaining a whole fork of google-java-format (which seems pretty gnarly)... realistically I think we're just gonna diverge.

My vote is that the upsides outweigh the downsides here. @CRogers @ferozco @carterkozak what do you think?

src/main/java/com/palantir/gradle/versions/FixLegacyJavaConfigurationsPlugin.java
dansanduleac: Do you dislike the concept or just the fact that it splits the `{@link`?

src/main/java/com/palantir/gradle/versions/FixLegacyJavaConfigurationsPlugin.java
```diff
@@ -88,26 +77,28 @@ private void injectVersions(Configuration conf, GetVersion getVersion) {
             }
             // Code adapted from:
             // https://github.com/nebula-plugins/nebula-dependency-recommender-plugin/blob/64ed7c6853f80b909918e6a595231a5e9803ae8b/src/main/groovy/netflix/nebula/dependency/recommender/DependencyRecommendationsPlugin.java
-            conf.getResolutionStrategy().eachDependency(details -> {
-                ModuleVersionSelector requested = details.getTarget();
+            conf.getResolutionStrategy()
+                    .eachDependency(details -> {
```
CRogers: Do you think this is fixable to move up to the previous line @dansanduleac?

src/main/java/com/palantir/gradle/versions/GetVersionPlugin.java
```diff
@@ -36,43 +36,44 @@
 
     @Override
     public void apply(Project project) {
-        project.getExtensions().getExtraProperties().set("getVersion", new Closure<String>(project, project) {
```
CRogers, line 39: I guess we need the "all on one line lambda" thing to work with anonymous classes too?

CRogers:
I think it's not perfect yet, but feel we can probably get it there, so am pro going down this approach further.

Another benefit is we get the "apply format to just the diff" feature, which means we can ensure excavators produce something sensible that passes checkstyle when rewriting code without forcing people to opt their entire codebase in.

src/main/java/com/palantir/gradle/versions/GetVersionPlugin.java
dansanduleac: yep! will collate all the nasty cases here and start addressing them once I fix the remaining two integration tests.

ferozco:
@dansanduleac for a given input AST does this formatter deterministically structure the code? Or is it like the eclipse formatter the produces a different structure based off the inputs format

src/main/java/com/palantir/gradle/versions/GradleWorkarounds.java
dansanduleac: I think this could be fixedby trying both "break initializer" (current behaviour) and newline after `=`.

Currently it will only allow a newline after `=` if the whole initializer fits on one line.

CRogers:
@ferozco It's the former, it creates an AST then throws away all the positional information, working on a stream on tokens, so for the same input ASTs you'll get the same output.

ferozco:
@iamdanfox I also worry about the maintenance cost but agree with your conclusion that the upside is worth it.

In particular, I'm  a huge fan of having the formatter deterministally produce formatted code since it take humans completely out of formatting decisions require less human intervention during roll out. Cost wise, g-j-f comes with both (intellij, eclipse) plugins in repo and so it should require minimal work to make it work with our variant of the formatter.

dansanduleac:
@uschi2000 hey - we're working on a fork of google-java-format that currently achieves this outcome.
Experimented with it this week, but since it already looks pretty neat, we'd like to move it open-source land. What are your thoughts?

src/main/java/com/palantir/gradle/versions/FixLegacyJavaConfigurationsPlugin.java
uschi2000: why wouldn't this render as `\nJavaPlugin.RUNTIME_CONFIGURATION_NAME`?

src/main/java/com/palantir/gradle/versions/GetVersionPlugin.java
```diff
@@ -105,30 +105,21 @@ private static String getVersion(Project project, String group, String name, Con
...
                 .map(mvi -> String.format("\t- %s:%s:%s", mvi.getGroup(), mvi.getName(), mvi.getVersion()))
                 .collect(Collectors.joining("\n"));
         return new GradleException(String.format(
                 "Unable to find '%s:%s' in %s. This may happen if you specify the version in versions.props but do not"
                         + " have a dependency in the configuration. The configuration contained:\n%s",
-                group,
-                name,
-                configuration,
-                actual));
+                group, name, configuration, actual));
```
uschi2000, line 114: what's the heuristic that pushes this into one line?

src/main/java/com/palantir/gradle/versions/VersionsProps.java
```diff
@@ -31,19 +31,16 @@
...
     private final Map<String, String> patternToPlatform;
 
     private VersionsProps(FuzzyPatternResolver fuzzyResolver) {
         this.fuzzyResolver = fuzzyResolver;
-        this.patternToPlatform = Sets
-                .difference(fuzzyResolver.versions().keySet(), fuzzyResolver.exactMatches())
-                .stream()
-                .collect(Collectors.toMap(key -> key, this::constructPlatform));
+        this.patternToPlatform =
+                Sets.difference(fuzzyResolver.versions().keySet(), fuzzyResolver.exactMatches()).stream()
```
uschi2000, line 42: this isn't great

src/main/java/com/palantir/gradle/versions/VersionsProps.java
```diff
@@ -56,9 +53,9 @@ public static VersionsProps loadFromFile(Path path) {
         FuzzyPatternResolver.Builder builder = FuzzyPatternResolver.builder();
         recommendations
                 .stringPropertyNames()
-                .forEach(name -> builder.putVersions(
-                        name.replaceAll("/", ":"),
-                        recommendations.getProperty(name).trim()));
+                .forEach(name -> builder.putVersions(name.replaceAll("/", ":"), recommendations
```
uschi2000: not great

src/main/java/com/palantir/gradle/versions/WhyDependencyTask.java
```diff
@@ -66,36 +66,37 @@ public final void fullLockState(Provider<FullLockState> provider) {
...
-            Optional<Dependents> entry = Stream
-                    .of(fullLockState.get().productionDeps(), fullLockState.get().testDeps())
-                    .map(state -> state.get(key))
-                    .filter(Objects::nonNull)
-                    .findFirst();
-            Dependents dependents =
-                    entry.orElseThrow(() -> new NullPointerException("Unable to find group/name in fullLockState"));
+                    Optional<Dependents> entry = Stream.of(fullLockState.get().productionDeps(), fullLockState
+                            .get()
+                            .testDeps())
```
uschi2000: not great

src/main/java/com/palantir/gradle/versions/lockstate/LockState.java
```diff
@@ -51,24 +49,19 @@
...
-                        Function.identity()));
+                        GradleComparators.MODULE_IDENTIFIER_COMPARATOR, Line::identifier, Function.identity()));
     }
 
     static LockState from(Stream<Line> productionLines, Stream<Line> testLines) {
-        return ImmutableLockState.of(
-                productionLines.collect(Collectors.toList()),
-                testLines.collect(Collectors.toList()));
+        return ImmutableLockState.of(productionLines.collect(Collectors.toList()), testLines.collect(Collectors
+                .toList()));
```
uschi2000: not great?

carterkozak:
I'm a bit worried about maintenance cost of a full fork, particularly around handling new java language features moving forward. Without continued investment we might hold ourselves back from upgrades, which are already a bit difficult to navigate.
If we're confident we can keep the maintenance cost minimal, and the output is significantly better than the eclipse formatter, I'm happy with it.
Caveat: Hard requirement that we support NON-NLS comments, which I think GJF claimed to support, but definitely does not work.

src/main/java/com/palantir/gradle/versions/lockstate/LockStates.java
```diff
@@ -66,12 +62,8 @@ private static Line componentWithDependentsToLine(ModuleVersionIdentifier compon
...
         HashCode hash = hasher.hash();
 
-        Line line = ImmutableLine.of(
-                component.getGroup(),
-                component.getName(),
-                component.getVersion(),
-                all.size(),
-                hash.toString());
+        Line line = ImmutableLine.of(component.getGroup(), component.getName(), component.getVersion(), all.size(), hash
+                .toString());
```
uschi2000: not great

dansanduleac:
@uschi2000 I should mention that this is not a finished product, just a proof of concept which currently only manages to enforce some of our code style.
I'd like to move further work directly to OSS before addressing the remaining cases that are not great.

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
dansanduleac: This should be easily fixable.

src/main/java/com/palantir/gradle/versions/FixLegacyJavaConfigurationsPlugin.java
dansanduleac: yep

src/main/java/com/palantir/gradle/versions/FixLegacyJavaConfigurationsPlugin.java
```diff
@@ -94,16 +84,16 @@ private void injectVersions(Configuration conf, GetVersion getVersion) {
...
                 for (ModuleVersionSelector force : conf.getResolutionStrategy().getForcedModules()) {
                     if (requested.getGroup().equals(force.getGroup()) && requested.getName().equals(force.getName())) {
-                        details.because(String.format(
-                                "Would have recommended a version for %s:%s, but a force is in place",
-                                requested.getGroup(),
-                                requested.getName()));
+                        details.because(
+                                String.format(
+                                        "Would have recommended a version for %s:%s, but a force is in place",
+                                        requested.getGroup(), requested.getName()));
```
dansanduleac: I made the non-splitting path a bit stricter so now if you have a single call inside this method (`details.because`) i.e. just `String.format()`, it won't try to split it anymore.
We can probably relax this later to allow some cases through.

src/main/java/com/palantir/gradle/versions/ConsistentVersionsPlugin.java
```diff
@@ -31,10 +31,9 @@ public final void apply(Project project) {
         project.getPluginManager().apply(GetVersionPlugin.class);
 
         project.allprojects(proj -> {
-            proj.getPluginManager()
-                    .withPlugin("java", plugin -> {
-                        proj.getPluginManager().apply(FixLegacyJavaConfigurationsPlugin.class);
-                    });
+            proj.getPluginManager().withPlugin("java", plugin -> {
+                proj.getPluginManager().apply(FixLegacyJavaConfigurationsPlugin.class);
+            });
```
dansanduleac, line 36: this is now fixed

src/main/java/com/palantir/gradle/versions/FixLegacyJavaConfigurationsPlugin.java
```diff
@@ -77,28 +78,26 @@ private void injectVersions(Configuration conf, GetVersion getVersion) {
             }
             // Code adapted from:
             // https://github.com/nebula-plugins/nebula-dependency-recommender-plugin/blob/64ed7c6853f80b909918e6a595231a5e9803ae8b/src/main/groovy/netflix/nebula/dependency/recommender/DependencyRecommendationsPlugin.java
-            conf.getResolutionStrategy()
-                    .eachDependency(details -> {
-                        ModuleVersionSelector requested = details.getTarget();
+            conf.getResolutionStrategy().eachDependency(details -> {
```
dansanduleac: this is also fixed now

src/main/java/com/palantir/gradle/versions/GetVersionPlugin.java
dansanduleac: GJF handles `String.format` arguments in a first class way.
It decouples the arguments from the format string, so the format string can be broken (split into lines) separately from the arguments.
Because the format string comes first, it gets broken (as it won't fit on one line) and that ensures another newline _after_ it, but then the arguments themselves fit on one line so they are left as is.

src/main/java/com/palantir/gradle/versions/FixLegacyJavaConfigurationsPlugin.java
carterkozak: I tend to optimize around putting the entire tag on a single line. Splitting on the space is as reasonable as anything, but I would prefer to avoid breaking them even if it leaves shorter lines before/after.
Sorry I missed your comment when you posted it!

dansanduleac: No problem, I think I prefer that too. Made a tracking issue and we can address once the more glaring formatting problems are solved.

src/main/java/com/palantir/gradle/versions/FixLegacyJavaConfigurationsPlugin.java
dansanduleac: Sorry I missed this. It won't ever do that, as the three parameters are all in a single "level", separated by breaks that are either taken all as a group or none at all.

As for why it accepts splitting there, that's due to an overly lenient last-argument-splitting heuristic that's now been fixed.
See https://github.com/palantir/gradle-consistent-versions/pull/305/files#diff-56e0cf074cb9bc301934d0fd45ed3508R56-R59

src/main/java/com/palantir/gradle/versions/GradleComparators.java
dansanduleac: Seems to have been fixed in the latest version: https://github.com/palantir/gradle-consistent-versions/blob/565fd9f3272a7ca7ba8f23d1e1ab22bd42f77911/src/main/java/com/palantir/gradle/versions/GradleComparators.java#L37-L41

src/main/java/com/palantir/gradle/versions/GradleWorkarounds.java
dansanduleac: Fixed thanks to https://github.com/palantir/palantir-java-format/pull/11, see https://github.com/palantir/gradle-consistent-versions/pull/305/files#diff-56e0cf074cb9bc301934d0fd45ed3508R56-R59

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
dansanduleac: Yes, it first considers `rootProject.getAllprojects()` a prefix (since it's a method on a field/variable), and then it further considers `rootProject.getAllprojects().stream()` to be a prefix because of the special casing you mentioned.

Therefore if this were to break, it would try to make it look like
```java
rootProject.getAllprojects()
        .stream()
```

before falling back to most splits if nothing else fits

```java
rootProject
        .getAllprojects()
        .stream()
```

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
dansanduleac: I think that works in the latest version.
For this code however `project.getConfigurations()` has been pulled onto the `=` line: https://github.com/palantir/gradle-consistent-versions/blob/565fd9f3272a7ca7ba8f23d1e1ab22bd42f77911/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java#L353-L357

dansanduleac:
Closing in favor of #305

PR #339, opened 2019-11-25
Clarify specificity of versions.props lines

dansanduleac:
## Before this PR

Unclear how multiple lines in `versions.props` matching the same jar work together, and which ones end up being used for that jar.

## After this PR
==COMMIT_MSG==
Clarify behaviour of multiple lines matching the same jar in `versions.props`.
==COMMIT_MSG==

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,9 +107,12 @@ junit:junit = 4.12
 org.assertj:* = 3.10.0
 ```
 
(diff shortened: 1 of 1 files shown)


ferozco (approved):

topherinternational:
This is very clear, thanks for the addition 🎉

PR #347, opened 2019-12-05
Fix extension/classifier dependencies with Gradle 6

dansanduleac:
## After this PR
==COMMIT_MSG==
 Fix bug in 1.13.0 with Gradle 6 where it would crash if you had dependencies with an explicit extension/classifier anywhere in your locked configurations.
==COMMIT_MSG==


src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -527,6 +519,7 @@ private void recursivelyCopyProjectDependenciesWithScope(
             Project currentProject,
             DependencySet dependencySet,
             Map<Configuration, String> copiedConfigurationsCache,
+            Map<ModuleIdentifier, GcvScope> scopes,
```
CRogers: Can we wrap this Map into it's own type? Then we can put low level the behaviour of setting a value with the comparator/getting it out again etc into that object.

dansanduleac: yep makes sense, done

CRogers (approved):

PR #365, opened 2019-12-19
add checkUnusedConstraints task

ferozco:
## Before this PR
versions.props files would slowly accumulate unused constraints as dependencies were added and removed from a project

## After this PR
==COMMIT_MSG==
Add `checkUnusedConstraints` task which will ensure that the set of constraints in the versions.props file is minimal. Calling the task with `--fix` will automatically remove any unnecessary constraints.
==COMMIT_MSG==

## Possible downsides?
I tried this out on a few repos and it seemed to do the right thing, there may be cases where it does not and so it will prevent people from merging.

We'll probably also want to update the automation so that the "fix" is automatically applied as part of the upgrade

### DIFF
--- a/src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsTask.java
+++ b/src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsTask.java
@@ -37,7 +37,6 @@ import org.gradle.api.artifacts.result.ResolutionResult;
 import org.gradle.api.file.RegularFile;
 import org.gradle.api.file.RegularFileProperty;
 import org.gradle.api.provider.Property;
-import org.gradle.api.provider.Provider;
 import org.gradle.api.provider.SetProperty;
 import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.InputFile;
@@ -73,10 +72,6 @@ public class CheckUnusedConstraintsTask extends DefaultTask {
 
     @Option(option = "fix", description = "Whether to apply the suggested fix to versions.props")
     public final void setShouldFix(boolean shouldFix) {
-        this.shouldFix.set(shouldFix);
-    }
-
-    final void setShouldFix(Provider<Boolean> shouldFix) {
         this.shouldFix.set(shouldFix);
     }
 
--- a/src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsTask.java
+++ b/src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsTask.java
@@ -92,10 +92,7 @@ public class CheckUnusedConstraintsTask extends DefaultTask {
 
         // assumes globs are sorted by specificity
         for (FuzzyPatternResolver.Glob glob : versionsProps.getFuzzyResolver().globs()) {
-            Optional<String> matchingArtifact = unmatchedArtifacts.stream().filter(glob::matches).findFirst();
-            if (matchingArtifact.isPresent()) {
-                unmatchedArtifacts.remove(matchingArtifact.get());
-            } else {
+            if (!unmatchedArtifacts.removeIf(glob::matches)) {
                 unusedConstraints.add(glob.getRawPattern());
             }
         }
--- a/src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsTask.java
+++ b/src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsTask.java
@@ -17,7 +17,6 @@
 package com.palantir.gradle.versions;
 
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Sets;
 import java.io.BufferedWriter;
 import java.io.File;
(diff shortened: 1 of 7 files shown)


src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsTask.java
```diff
@@ -0,0 +1,163 @@
...
+        Set<String> artifacts = getClasspath().get();
+        VersionsProps versionsProps = VersionsProps.loadFromFile(getPropsFile().get().getAsFile().toPath());
+
+        Set<String> exactConstraints = versionsProps.getFuzzyResolver().exactMatches();
+        Stream<String> unusedExactConstraints = Sets.difference(exactConstraints, artifacts).stream();
+
+        Sets.SetView<String> unmatchedArtifacts = Sets.difference(artifacts, exactConstraints);
+        Stream<String> unusedGlobConstraints = versionsProps.getFuzzyResolver().globs().stream()
+                .filter(glob -> !unmatchedArtifacts.stream().anyMatch(glob::matches))
+                .map(FuzzyPatternResolver.Glob::getRawPattern);
```
dansanduleac: the logic here should be more like
- start with all globs
- for each unmatchedArtifacts:
	- find the _most specific_ glob that matches (there should be a method for this, which essentially counts the number of non-`*` characters in the glob)
	- remove it from the set

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsTask.java
```diff
@@ -0,0 +1,167 @@
...
+
+        // assumes globs are sorted by specificity
+        for (FuzzyPatternResolver.Glob glob : versionsProps.getFuzzyResolver().globs()) {
+            Optional<String> matchingArtifact = unmatchedArtifacts.stream().filter(glob::matches).findFirst();
+            if (matchingArtifact.isPresent()) {
+                unmatchedArtifacts.remove(matchingArtifact.get());
+            } else {
+                unusedConstraints.add(glob.getRawPattern());
+            }
+        }
```
dansanduleac, line 92: This is not quite what I meant.
The above code 'validates' any glob that matches at least an artifact (then removing the artifact from the pool).
But a glob could match multiple artifacts, rendering other lower-priority globs unused, even if they also match some artifacts. In the code above, a second glob could come in and steal a 2nd artifact that should have matched the first glob.

dansanduleac: Actually this functionality is already on FuzzyResolver: 
```java
unmatchedArtifacts.stream().flatMap(ua -> Stream.stream(versionsProps.getFuzzyResolver().patternFor(ua)))
```
gives you a list of glob.rawPattern that did match stuff, then you just need to subtract this from unusedConstraints.
Edit: assuming you start with `unusedConstraints = <all constraints> - <artifacts>` not `<exactConstraints> - <artifacts>` like it is now.

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsTask.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    @Option(option = "fix", description = "Whether to apply the suggested fix to versions.props")
+    public final void setShouldFix(boolean shouldFix) {
+        this.shouldFix.set(shouldFix);
+    }
+
+    final void setShouldFix(Provider<Boolean> shouldFix) {
+        this.shouldFix.set(shouldFix);
+    }
```
dansanduleac: this is unused, do we need it?

dansanduleac:
👍 , wanna mention this task in the readme as well?

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsTask.java
ferozco: done

dweiss:
I'm working on a gradle build for Apache Lucene and we're getting false hits with this patch. checkUnusedConstraints fails check and reports an extra dependency, --fix removes it. 
```
git clone https://github.com/apache/lucene-solr -b gradle-master --depth 1
cd lucene-solr
git checkout 0166e89dba1359f443067750789aae36c2d6d35c
./gradlew checkUnusedConstraints
```
reports:
```
> There are unused pins in your versions.props:
  [com.fasterxml.woodstox:*, org.apache.commons:commons-collections4:4.2, info.ganglia.gmetric4j:gmetric4j, org.apache.pdfbox:jbig2-imageio]
```

But at least one of these dependencies is actually used in a submodule (explicitly):
https://github.com/apache/lucene-solr/blob/gradle-master/solr/contrib/extraction/build.gradle#L34

and when removed the build fails:
```
./gradlew verifyLocks
> Task :verifyLocks FAILED

* What went wrong:
> Could not compute lock state from configuration 'unifiedClasspath' due to unresolved dependencies:
   * org.apache.commons:commons-collections4 (requested: 'org.apache.commons:commons-collections4' because: requested)
        Failures:
           - Could not find org.apache.commons:commons-collections4:.
```

PR #375, opened 2020-01-09
Ensure that entries in versions.props file are well formed

### LINKED ISSUE #373: More aggressively validate versions.props
## What happened?

https://github.com/palantir/gradle-consistent-versions/issues/372 showcased that we allow multiple ways of defining constraints in `versions.props`:

- `group:name = version`
- `group:name:version`

While the latter worked before, it doesn't get parsed correctly by the new `checkUnusedConstraints` task, leading to false positives.

Enforce that `versions.props` only contains constraints of the form `group:name = version` in order to simplify the logic that reads and validates that file.

ferozco:
Closes #373 
## Before this PR
Users could inadvertently specify constraints in way that could result in confusing errors when checking for unused constraints.

ex:
```
com.palantir:test:1.0.0
// Some comment
```

## After this PR
==COMMIT_MSG==
Validate that versions.props entries are well formed
==COMMIT_MSG==

## Possible downsides?
N/A


build.gradle
```diff
@@ -54,6 +54,8 @@ dependencies {
     testImplementation 'com.netflix.nebula:nebula-test'
     testImplementation 'com.fasterxml.jackson.core:jackson-databind'
     testImplementation "org.junit.jupiter:junit-jupiter-api"
+    testImplementation 'org.junit.jupiter:junit-jupiter'
```
dansanduleac:
```suggestion
    testRuntimeOnly 'org.junit.jupiter:junit-jupiter'
```

dansanduleac (approved):

ferozco:
👍

PR #396, opened 2020-01-27
Fail on duplicate constraints

ferozco:
## Before this PR
We would silently ignore duplicate constraints preferring the last constraint encountered.

## After this PR
==COMMIT_MSG==
Fail to parse a versions.props file if there are duplicate constraints
==COMMIT_MSG==

## Possible downsides?
There might be some cases where users had duplicate constraints and so this will require manual action


dansanduleac:
👍

PR #343, opened 2019-12-04
Gradle 6.0 support

CRogers:
## Before this PR
Running GCV on Gradle 6.0 results in this error message:

```
> Failed to apply plugin [class 'com.palantir.gradle.versions.VersionsLockPlugin']
   > Cannot add attributes or capabilities on a dependency that specifies artifacts or configuration information
```

## After this PR
==COMMIT_MSG==
`gradle-consistent-versions` now supports Gradle 6.0. The minimum required Gradle version has been raised to 5.3
==COMMIT_MSG==

## Possible downsides?
Minimum Gradle version needs to be bumped to support usage of capabilities.

PR #413, opened 2020-02-07
Delete dead code

iamdanfox:
Seems like we aren't using this anywhere? Minimum claims to be 5.3


dansanduleac:
The comment mentions that one of the fixes only made it in 5.6-rc-1 though, while the minimum is 5.3 as you say.

dansanduleac:
Actually I think since #343 we no longer use this code path in the way that was broken until 5.6-rc-1 so it should be fine to delete this now 👍

dansanduleac:
🐞

PR #418, opened 2020-02-12
Create gradle-wrapper-validation.yml

dansanduleac:
## After this PR
==COMMIT_MSG==

==COMMIT_MSG==

### DIFF
--- a/.github/workflows/gradle-wrapper-validation.yml
+++ b/.github/workflows/gradle-wrapper-validation.yml
@@ -1,5 +1,5 @@
 name: "Validate Gradle Wrapper"
-on: [push, pull_request]
+on: [push]
 
 jobs:
   validation:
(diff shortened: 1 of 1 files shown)


.github/workflows/gradle-wrapper-validation.yml
```diff
@@ -0,0 +1,10 @@
+name: "Validate Gradle Wrapper"
+on: [push, pull_request]
```
iamdanfox: not sure what the benefit of having both is here - maybe just `push`?

dansanduleac: Yep I agree, was just following their [example](https://github.com/gradle/wrapper-validation-action#add-a-new-dedicated-workflow) but it seems kinda pointless if it runs twice.

image

PR #419, opened 2020-02-12
Don't try to resolve legacy java configurations

dansanduleac:
## Before this PR

`checkUnusedConfigurations` would sometimes trigger a gradle bug (https://github.com/gradle/gradle/issues/11844) on resolving certain java configurations (`compile`, `runtime`, `compileOnly`) that are only resolvable for backcompat reasons but for all intents and purposes should not be resolved.

## After this PR
==COMMIT_MSG==
Fix regression in 1.14.0 so that checkUnusedConstraints stops trying to resolve legacy java configurations.
==COMMIT_MSG==

### DIFF
--- a/src/main/java/com/palantir/gradle/versions/GradleConfigurations.java
+++ b/src/main/java/com/palantir/gradle/versions/GradleConfigurations.java
@@ -1,0 +1,56 @@
+/*
+ * (c) Copyright 2020 Palantir Technologies Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


ferozco (approved):

src/main/java/com/palantir/gradle/versions/GradleConfigurations.java
```diff
@@ -0,0 +1,56 @@
...
+
+final class GradleConfigurations {
+    /**
+     * Filters out both the unresolvable configurations but also the legacy java configurations that should not be
+     * resolved.
+     */
+    public static Stream<Configuration> getResolvableConfigurations(Project project) {
+        return project.getConfigurations().stream()
+                .filter(Configuration::isCanBeResolved)
+                .filter(conf -> !getLegacyJavaConfigurations(project).contains(conf.getName()));
```
schlosna, line 35: Should we pull this set out to a local variable captured by lambda to avoid computing it for each configuration?

Issue #423, opened 2020-02-14
Cannot change dependencies of dependency configuration 'foo'


kristian-elder:
## What happened?

We're trialling the use of the consistent versions plugin on our project, and when it works, it works brillantly. However, with certain combinations of multi project gradle builds, we get an exception when trying to build the project. Whether or not the exception happens appear to be dependent on the specific set of modules and their naming - it's possible that it is triggered by some internal iteration order of modules.

When running a `./gradlew clean build` at the root of a multi module gradle build containing dependencies between the modules themselves, the build fails with `Caused by: org.gradle.api.InvalidUserDataException: Cannot change dependencies of dependency configuration ':<redacted>:apiElements' after it has been included in dependency resolution.` (see full stack trace below).

This happens under gradle 6.1, consistent versions plugin 0.17.2, and a Java 11 runtime (though it seems to fail also with other combinatins).

## What did you want to happen?

For the gradle build to work.

## Full stack trace of error:

```
org.gradle.api.ProjectConfigurationException: A problem occurred configuring root project '<redacted>'.
        at org.gradle.configuration.project.LifecycleProjectEvaluator.wrapException(LifecycleProjectEvaluator.java:80)
        at org.gradle.configuration.project.LifecycleProjectEvaluator.addConfigurationFailure(LifecycleProjectEvaluator.java:73)
        at org.gradle.configuration.project.LifecycleProjectEvaluator.access$600(LifecycleProjectEvaluator.java:53)
        at org.gradle.configuration.project.LifecycleProjectEvaluator$NotifyAfterEvaluate.run(LifecycleProjectEvaluator.java:199)
        at org.gradle.internal.operations.DefaultBuildOperationExecutor$RunnableBuildOperationWorker.execute(DefaultBuildOperationExecutor.java:402)
        at org.gradle.internal.operations.DefaultBuildOperationExecutor$RunnableBuildOperationWorker.execute(DefaultBuildOperationExecutor.java:394)
        at org.gradle.internal.operations.DefaultBuildOperationExecutor$1.execute(DefaultBuildOperationExecutor.java:165)
        at org.gradle.internal.operations.DefaultBuildOperationExecutor.execute(DefaultBuildOperationExecutor.java:250)
        at org.gradle.internal.operations.DefaultBuildOperationExecutor.execute(DefaultBuildOperationExecutor.java:158)
[...]

PR #424, opened 2020-02-14
Fail fast if outgoing configurations are realized early

dansanduleac:
## Before this PR

Hard to debug bad interactions with other plugins.
For instance, the kotlin plugin realizes the dependencies of a bunch of java configurations early (at configuration time), and GCV doesn't allow that (see https://github.com/palantir/gradle-consistent-versions/issues/423#issuecomment-586383644).

However, by the time GCV fails, it will fail like so, and the stack trace won't point to who "included this configuration in dependency resolution" in the first place.

```
org.gradle.api.InvalidUserDataException: Cannot change dependencies of dependency configuration ':<redacted>:apiElements' after it has been included in dependency resolution.
```

## After this PR
==COMMIT_MSG==
Immediately throw an error capturing the stacktrace when other plugins / gradle code are attempting to realize configuration dependencies/constraints early (at configuration-time) and throw an error.
==COMMIT_MSG==

### DIFF
--- a/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
@@ -227,6 +227,21 @@ public class VersionsLockPlugin implements Plugin<Project> {
         // (but that's internal)
         project.getPluginManager().apply("java-base");
 
+        // This is helpful to short-circuit bad interactions with other plugins and pinpoint which code path caused
+        // the error. The goal is to hook up guard rails against the same configurations that are lazily configured by
+        // configurePublishConstraints (which usually happens too late to produce a good stack-trace: in the root
+        // project's afterEvaluate).
+        // Ideally, this code would go happen together with that function, and be wired up to compute the locked
+        // configurations lazily, rather than eagerly but inside rootProject.afterEvaluate which happens very late.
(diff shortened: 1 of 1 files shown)


src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -289,6 +304,23 @@ public final void apply(Project project) {
...
+                GradleWorkarounds.fixListProperty(project.getObjects().listProperty(DependencyConstraint.class));
+        // Guard against bad plugins.
+        constraintsProperty.addAll(project.provider(() -> {
+            if (GradleWorkarounds.isConfiguring(project.getRootProject().getState())) {
+                throw new RuntimeException(String.format(
+                        "A plugin has realized the dependencies of %s while the project is still configuring. "
+                                + "This is not allowed when using GCV.",
+                        configuration.get()));
+            }
+            return ImmutableList.of();
```
fawind, line 319: We need to return this so the resolution constraint gets triggered in `addAllLater`?

dansanduleac: yep, I didn't know how else to add this as a "lazy effect" essentially

fawind (approved):

PR #428, opened 2020-02-19
Fix io.grpc dependency inconsistencies 

### LINKED ISSUE #412: Inconsistencies between local & CI machines grpc [1.23.0]
## What happened?

```

* What went wrong:
> Found dependencies whose dependents changed:
  -io.grpc:grpc-api:1.23.0 (5 constraints: 02360e50)
  +io.grpc:grpc-api:1.23.0 (5 constraints: 4a352adc)
  -io.grpc:grpc-core:1.23.0 (1 constraints: 7f098fa3)
  +io.grpc:grpc-core:1.23.0 (1 constraints: c708ee8f)
  -io.netty:netty-codec-http2:4.1.38.Final (1 constraints: a10b58e3)
  +io.netty:netty-codec-http2:4.1.38.Final (1 constraints: e90a73cb)
  
  Please run './gradlew --write-locks'.
```

Locally (mac):
```
$ ./gradlew why --hash 7f098fa3

> Task :why
io.grpc:grpc-core:1.23.0
        io.grpc:grpc-netty -> [1.23.0]
```

This square bracket version seems to be what's in the upstream POM: https://repo1.maven.org/maven2/io/grpc/grpc-netty/1.23.0/grpc-netty-1.23.0.pom, and it seems like the grpc repo [intentionally sets this](https://github.com/grpc/grpc-java/blob/557da62305b7b4499300f9b201a25b44fa0d60ef/build.gradle#L374).

On CI (linux)
```
$ ./gradlew why --hash c708ee8f

> Task :why
io.grpc:grpc-core:1.23.0
	io.grpc:grpc-netty -> 1.23.0
```

This repros with java 8 and 11.

Running `find ~/.gradle/caches/ -path '*grpc*' -delete` seems to solve this sometimes?

Local and CI should always be consistent. Unsure if this is a gradle bug, or if the upstream artifacts are actually mutable somehow??

CRogers:
## Before this PR
Fixes #412. Thanks to a [likely Gradle bug](https://github.com/gradle/gradle/issues/12229), the generated lock file can be different on different machines/JVMs. This is particularly bad for `io.grpc` dependencies as they seem to be the only commonly used dependencies that declare version ranges like `[1.27.1]` for just one specific version, which triggers the Gradle bug:

```
> Found dependencies whose dependents changed:
  -io.grpc:grpc-api:1.23.0 (5 constraints: 02360e50)
  +io.grpc:grpc-api:1.23.0 (5 constraints: 4a352adc)
  -io.grpc:grpc-core:1.23.0 (1 constraints: 7f098fa3)
  +io.grpc:grpc-core:1.23.0 (1 constraints: c708ee8f)
  -io.netty:netty-codec-http2:4.1.38.Final (1 constraints: a10b58e3)
  +io.netty:netty-codec-http2:4.1.38.Final (1 constraints: e90a73cb)
  
  Please run './gradlew --write-locks'.
```

## After this PR
==COMMIT_MSG==
`io.grpc` dependencies should no longer cause `versions.lock` generation inconsistencies between different machines.
==COMMIT_MSG==

We do this by just changing the single version range `[1.27.1]` to simply `1.27.1` before hashing them.

## Possible downsides?
Feels sad to work around this instead of fixing the actual Gradle bug - doing so simply hides the issue for longer. We must follow up with the Gradle team as well.


src/test/groovy/com/palantir/gradle/versions/lockstate/LockStatesTest.java
```diff
@@ -0,0 +1,70 @@
...
+        assertThat(LockStates.prettyPrintConstraints(Dependents.of(dependents)))
+                .containsExactly(
+                        "io.grpc:grpc-api -> 1.27.1",
+                        "io.grpc:grpc-core -> 1.27.1",
+                        "io.grpc:grpc-netty -> {1.27.1, 1.27.1}",
+                        "something:else -> [1.27.1]");
+    }
+
+    private ComponentIdentifier componentIdentifier(String componentIdentifier) {
+        ComponentIdentifier grpcApi = mock(ComponentIdentifier.class);
```
CRogers, line 60: It feels kinda bad to mock these types out, since we don't own them, but the alternative is using internal Gradle types. I think it's ok since we essentially just use `toString`/`getDisplayString` on them rather than mocking out any real complicated behaviour.

src/main/java/com/palantir/gradle/versions/lockstate/LockStates.java
```diff
@@ -99,6 +102,16 @@ private static Line componentWithDependentsToLine(ModuleVersionIdentifier compon
                 .collect(toList());
     }
 
+    private static String versionConstraintToString(VersionConstraint versionConstraint) {
+        String constraintString = versionConstraint.toString();
+
+        if (SINGLE_VERSION_RANGE.matcher(constraintString).matches()) {
```
dansanduleac, line 108: does this match the whole string necessarily?

CRogers: `.matches()` does yes. (`.find()` doesn't):

> Attempts to match the entire region against the pattern.

dansanduleac (approved):

dansanduleac:
helloooooo

PR #476, opened 2020-04-15
Remove nebula dependency

ferozco:
## Before this PR
We pulled in nebula (and transitively the kotlin runtime) just so we could verify the recommendation strategy

## After this PR
Change the dependency to be compileOnly and guard the access so that we're guaranteed to have the classes on the classpath if we execute the code.

## Possible downsides?
N/A

### DIFF
--- a/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
@@ -56,8 +56,6 @@ import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import javax.inject.Inject;
-import netflix.nebula.dependency.recommender.RecommendationStrategies;
-import netflix.nebula.dependency.recommender.provider.RecommendationProviderContainer;
 import org.gradle.api.GradleException;
 import org.gradle.api.Named;
 import org.gradle.api.NamedDomainObjectProvider;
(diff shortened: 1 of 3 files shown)


src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -432,14 +430,7 @@ private static void checkPreconditions(Project project) {
...
                 sub.getPluginManager().withPlugin("nebula.dependency-recommender", plugin -> {
-                    RecommendationProviderContainer container =
-                            sub.getExtensions().findByType(RecommendationProviderContainer.class);
-                    if (container.getStrategy() == RecommendationStrategies.OverrideTransitives) {
-                        throw new GradleException("Must not use strategy OverrideTransitives for "
-                                + sub
-                                + ". "
-                                + "Use this instead: dependencyRecommendations { strategy ConflictResolved }");
-                    }
+                    NebulaUtils.verifyRecommendationStrategy(sub);
```
dansanduleac, line 433: this won't work if nebula is brought in via `plugins{}` i.e. in a different classloader.

dansanduleac (commented):
I'm supportive of removing nebula from GCV, though I think instead of still half-supporting it we should probably just ban it altogether.
It's almost always a mistake to have both of them on the classpath.

dansanduleac:
We probably need to fix up some tests too, so that they bring in nebula rather than expecting it to be there.

ferozco:
Apologies, this fell off of my radar. I'll try and fix it up

PR #489, opened 2020-05-21
relax constraint for configure-on-demand

bjlaub:
This change relaxes the precondition constraint that checks that
configure-on-demand is disabled. We now only require it to be disabled
when running the "verifyLocks" task, or when invoked with
"--write-locks".

## Before this PR
Previously GCV would throw when the plugin was applied to any project that had set `org.gradle.configureondemand=true`, but this seems overly restrictive. The constraint should be necessary only when verifying or updating the lockfile, which requires configuring all projects.

## After this PR
==COMMIT_MSG==
This change relaxes the precondition constraint that checks that
configure-on-demand is disabled. We now only require it to be disabled
when running the "verifyLocks" task, or when invoked with
"--write-locks".

This gives projects more flexibility to flag on configure-on-demand in certain scenarios where it might make sense.
==COMMIT_MSG==

## Possible downsides?
Enabling "dual-mode" configure-on-demand (e.g. disabled when writing/verifying locks, and enabled otherwise) could be tricky, since it can only be enabled/disabled via the gradle property or command line switch. That means in order for users to get the best of both worlds that this change tries to address, they'll likely have to do some tricks with the gradle wrapper script.

### DIFF
--- a/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
@@ -434,10 +434,13 @@ public class VersionsLockPlugin implements Plugin<Project> {
             throw new GradleException("Must be applied only to root project");
         }
 
-        Preconditions.checkState(
-                !project.getGradle().getStartParameter().isConfigureOnDemand(),
-                "Gradle Consistent Versions doesn't currently work with configure-on-demand, please remove"
-                        + " 'org.gradle.configureondemand' from your gradle.properties");
+        // check that configure-on-demand is false when running with --write-locks
+        if (project.getGradle().getStartParameter().isWriteDependencyLocks()) {
+            Preconditions.checkState(
+                    !project.getGradle().getStartParameter().isConfigureOnDemand(),
+                    "configure-on-demand cannot be enabled when gradle is invoked with --write-locks;"
+                            + " please remove 'org.gradle.configureondemand' from your gradle.properties");
+        }
 
         Multimap<String, Project> coordinateDuplicates = LinkedHashMultimap.create();
         Set<Project> subprojectsLeft = new HashSet<>(project.getSubprojects());
--- a/changelog/@unreleased/pr-489.v2.yml
+++ b/changelog/@unreleased/pr-489.v2.yml
@@ -1,0 +1,11 @@
+type: improvement
+improvement:
+  description: This change relaxes the precondition constraint that checks that
+    configure-on-demand is disabled. We now only require it to be disabled
+    when running the "verifyLocks" task, or when invoked with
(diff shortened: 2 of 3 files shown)


src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -434,10 +434,13 @@ private static void checkPreconditions(Project project) {
...
-        Preconditions.checkState(
-                !project.getGradle().getStartParameter().isConfigureOnDemand(),
-                "Gradle Consistent Versions doesn't currently work with configure-on-demand, please remove"
-                        + " 'org.gradle.configureondemand' from your gradle.properties");
+        // check that configure-on-demand is false when running with --write-locks
+        if (project.getGradle().getStartParameter().isWriteDependencyLocks()) {
+            Preconditions.checkState(
+                    !project.getGradle().getStartParameter().isConfigureOnDemand(),
+                    "configure-on-demand cannot be enabled when gradle is invoked with --write-locks;"
+                            + " please remove 'org.gradle.configureondemand' from your gradle.properties");
```
dansanduleac, line 442: maybe pull this out to a package-private function?

dansanduleac: ah nvm the message is different

bjlaub: 🤷 i'd be happy to do that and modify the message thusly:
```
configure-on-demand cannot be enabled when gradle is invoked with '--write-locks' or when running the 'verifyLocks' task
```
if you prefer?

dansanduleac: up to you, i'm fine either way. shall we add a changelog?

dansanduleac:
👍

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
bjlaub: @dansanduleac doesn't look like i have permissions to add via changelog-bot? i can't click the button, apparently...

carterkozak: I don't think it knows how to push to a fork unfortunately. I thought we gave everyone basic perms to push to these repos.

Might be easiest to write the changelog manually, like a cave-person.

bjlaub: ah doh... ok

bjlaub: alrighty, added manually

PR #491, opened 2020-05-25
Support dependency cycles when determining scope

ferozco:
## Before this PR
We would cause a stack overflow when attempting to determine the scope of a given dependency. This was caused because we would perform a DFS without short circuiting to find direct dependencies.

Example exception:
```
Caused by: java.lang.StackOverflowError
        at com.google.common.collect.RegularImmutableMap.get(RegularImmutableMap.java:122)
        at com.palantir.gradle.versions.DirectDependencyScopes.getScopeFor(DirectDependencyScopes.java:40)
        at com.palantir.gradle.versions.VersionsLockPlugin.lambda$getScopeRecursively$46(VersionsLockPlugin.java:780)
        at com.palantir.gradle.versions.VersionsLockPlugin.getScopeRecursively(VersionsLockPlugin.java:786)
        at com.palantir.gradle.versions.VersionsLockPlugin.lambda$getScopeRecursively$46(VersionsLockPlugin.java:784)
        at com.palantir.gradle.versions.VersionsLockPlugin.getScopeRecursively(VersionsLockPlugin.java:786)
```

## After this PR
==COMMIT_MSG==
Properly handle cycles in the dependency graph when determining scope
==COMMIT_MSG==

TODO:
 - still need to investigate why we are discovering a different scope for dependencies in tests

## Possible downsides?
N/A

### DIFF
--- a/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
@@ -797,7 +797,7 @@ public class VersionsLockPlugin implements Plugin<Project> {
             // If the dependency came from a project, then the requested ModuleIdentifier should be in the
             // edgeScopes. Otherwise, recurse until we find a project dependent.
             Optional<GcvScope> maybeScope = directDependencyScopes.getScopeFor(requestedModule);
-            if (dependent.getFrom().getId() instanceof ProjectComponentIdentifier && maybeScope.isPresent()) {
+            if (dependent.getSelected().getId() instanceof ProjectComponentIdentifier && maybeScope.isPresent()) {
                 discoveredScopes.add(maybeScope.get());
                 continue;
             }
@@ -809,6 +809,8 @@ public class VersionsLockPlugin implements Plugin<Project> {
         GcvScope scope = discoveredScopes.stream()
                 .min(GCV_SCOPE_COMPARATOR)
                 .orElseThrow(() -> new RuntimeException("Couldn't determine scope for dependency: " + component));
+
+        scopeCache.put(component, scope);
         for (ResolvedDependencyResult traversedComponent : traversedComponents) {
             log.warn("updating transitive {} {}", component, traversedComponent);
             scopeCache.put(traversedComponent.getFrom(), scope);
--- a/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
@@ -40,8 +40,10 @@ import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.ArrayDeque;
 import java.util.Collection;
 import java.util.Comparator;
+import java.util.Deque;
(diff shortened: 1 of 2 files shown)


src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -769,25 +772,49 @@ private static GcvScope getScopeRecursively(
...
+        Set<GcvScope> discoveredScopes = new HashSet<>(2);
+        while (!stack.isEmpty()) {
+            ResolvedDependencyResult dependent = stack.removeFirst();
+            if (dependent.isConstraint()
+                    || !(dependent.getRequested() instanceof ModuleComponentSelector)
+                    || traversedComponents.contains(dependent)) {
+                continue;
+            }
+
+            Optional<GcvScope> cachedValue = Optional.ofNullable(scopeCache.get(dependent.getFrom()));
```
dansanduleac, line 789: `dependent.getFrom()` is not the same as `((ModuleComponentSelector) dependent.getRequested()).getModuleIdentifier()`

for a dependency `org:foo:1.0 -> org:bar:1.1`
* `.from` returns the `ResolvedComponentResult` `org:foo:1.0`
* `.requested.moduleIdentifier` returns the `ModuleIdentifier` `org:bar`

so you're putting the wrong type into scopeCache.get

dansanduleac: orr.. never mind, I got myself confused with these types and what was happening before, it actually makes sense so far

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -769,25 +772,49 @@ private static GcvScope getScopeRecursively(
...
+            if (cachedValue.isPresent()) {
+                discoveredScopes.add(cachedValue.get());
+                continue;
+            }
```
dansanduleac, line 793: Previously this was checking that a component i.e. a `ResolvedComponentResult` (whose dependents we later recurse through) is in the cache.

That component was previously coming from `dependent.getFrom()` in the recursive call, but now we're discarding that `dependent.getFrom()`, instead dumping its *dependents* into the stack.

However, from the point of view of the _dependents_ (arrows going _into_ the component), the component you got them from is actually `dependent.getSelected()`

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -769,25 +772,51 @@ private static GcvScope getScopeRecursively(
...
+
+        GcvScope scope = discoveredScopes.stream()
                 .min(GCV_SCOPE_COMPARATOR)
                 .orElseThrow(() -> new RuntimeException("Couldn't determine scope for dependency: " + component));
 
-        scopeCache.put(component, gcvScope);
-        return gcvScope;
+        scopeCache.put(component, scope);
+        for (ResolvedDependencyResult traversedComponent : traversedComponents) {
+            log.warn("updating transitive {} {}", component, traversedComponent);
```
dansanduleac: probably don't wanna leave this as warn

dansanduleac:
👍

PR #522, opened 2020-06-23
Check unused constraints can be up-to-date

j-baker:
This task takes 1.8 seconds for me, and only is relevant in a few cases (when I
change my props file, when I change my classpath). Here, we enable the
task to be up-to-date more frequently.


carterkozak (approved):
Thanks

PR #537, opened 2020-07-08
Improvement: Support cycles in dependency graph

ferozco:
## Before this PR
We had previously attempted to support cycles #491, but had introduced a subtle bug that caused us to revert the change.

## After this PR
==COMMIT_MSG==
Support cycles in dependency graph
==COMMIT_MSG==

## Possible downsides?
N/A

### DIFF
--- a/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
@@ -40,8 +40,10 @@ import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.ArrayDeque;
 import java.util.Collection;
 import java.util.Comparator;
+import java.util.Deque;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -77,6 +79,7 @@ import org.gradle.api.artifacts.component.ProjectComponentIdentifier;
 import org.gradle.api.artifacts.dsl.DependencyConstraintHandler;
 import org.gradle.api.artifacts.result.ResolutionResult;
 import org.gradle.api.artifacts.result.ResolvedComponentResult;
+import org.gradle.api.artifacts.result.ResolvedDependencyResult;
 import org.gradle.api.artifacts.result.UnresolvedDependencyResult;
 import org.gradle.api.attributes.Attribute;
 import org.gradle.api.attributes.AttributesSchema;
@@ -701,7 +704,7 @@ public class VersionsLockPlugin implements Plugin<Project> {
         resolutionResult.getAllComponents().stream()
                 .filter(component -> component.getId() instanceof ModuleComponentIdentifier)
                 .forEach(component -> {
-                    GcvScope scope = getScopeRecursively(component, scopeCache, directDependencyScopes);
+                    GcvScope scope = getScope(component, scopeCache, directDependencyScopes);
                     switch (scope) {
                         case PRODUCTION:
                             builder.putProductionDeps(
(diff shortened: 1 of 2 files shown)


src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -740,18 +740,14 @@ private static GcvScope getScope(
...
-            if (dependent.isConstraint()
-                    || !(dependent.getRequested() instanceof ModuleComponentSelector)
-                    || traversedComponents.contains(dependent)) {
+            if (dependent.isConstraint() || !(dependent.getRequested() instanceof ModuleComponentSelector)) {
                 continue;
-            }
-
-            Optional<GcvScope> cachedValue = Optional.ofNullable(scopeCache.get(dependent.getFrom()));
-            if (cachedValue.isPresent()) {
-                discoveredScopes.add(cachedValue.get());
```
dansanduleac: why did we remove this? it means we're never relying on the cache anymore except for direct hits.

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -740,18 +740,14 @@ private static GcvScope getScope(
...
+            } else if (traversedComponents.contains(dependent)) {
                 continue;
             }
 
+            traversedComponents.add(dependent);
```
dansanduleac, line 749: I take it this was the bug, we could be continuing before we add this dependent to traversedComponents.

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -762,18 +758,14 @@ private static GcvScope getScope(
...
         }
 
         GcvScope scope = discoveredScopes.stream()
                 .min(GCV_SCOPE_COMPARATOR)
                 .orElseThrow(() -> new RuntimeException("Couldn't determine scope for dependency: " + component));
 
         scopeCache.put(component, scope);
-        for (ResolvedDependencyResult traversedComponent : traversedComponents) {
-            scopeCache.put(traversedComponent.getSelected(), scope);
-        }
```
dansanduleac: Coudln't the removal of this trigger a pathological N^2 case if getScope is called on the leaves first and so on going up the graph?
Even if that's not necessarily the case, wouldn't want to rely on implicit gradle behaviour as we know they change things like this all the time.

dansanduleac:
It seems to be that we're complicating the code a bit too much by making it iterative.
We've introduced some bugs in the first attempt, and the current version keeps only leaf-node caching it seems like (not caching the nodes on the path from root to node being queried), which I don't think it's safe to do / might be slow on very deep graphs.

What if we just change the original implementation to return `Optional.empty` if the node is already in the set of visited nodes, and skip recursing into those dependents?

dansanduleac:
@robert3005 what do you think - should we just revert to the simpler recursion or do you think it's fine? I'm worried it's a bit harder to reason about now, even though it probably works.

robert3005:
I don't think this is that much more complicated than recurisve implementation. I mostly care about it since we manage to hit this in some rare edge cases and then you're just stuck.

dansanduleac:
Alright, well I don't feel strongly about that then, except I think we removed some caching that we might want - https://github.com/palantir/gradle-consistent-versions/pull/537#discussion_r453627422 (link is broken, i pointed to this code - https://github.com/palantir/gradle-consistent-versions/pull/537/commits/5cf0cc410b980970c66fc1dda087f707b95d4cc9#diff-2382390745b175d9f349bda21607d522R749-R753)

robert3005:
I have added that code back

dansanduleac:
I guess ideally we'd have a test for the case where it failed last time, but we can roll with it for now. 👍

robert3005:
I think the case where we failed were only on two internal projects and one of them was some experiment that @ferozco and I have been working on. Not sure we can easily repro this here.

PR #557, opened 2020-08-12
Enforce lock constraints using shared platform

dansanduleac:
## Before this PR

Creating lots of constraints into every single configuration being resolved (even though technically only once per project, after which they get inherited by all configurations).

This can lead to super large build scans (300MB for a large internal monorepo) where constraints are needlessly duplicated.

## After this PR
==COMMIT_MSG==
Manage lock file constraints using a single platform, reducing the number of constraints that have to be created.

This works in a similar way to [Java Platform Plugin](https://docs.gradle.org/current/userguide/java_platform_plugin.html). We can't just use that plugin directly, as it makes too many assumptions about the structure of the build where it's applied (e.g. needs its own project that isn't otherwise a java project, to hold the platform) that we can't make in GCV.
==COMMIT_MSG==

### DIFF
--- a/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
@@ -839,22 +839,16 @@ public class VersionsLockPlugin implements Plugin<Project> {
 
     private static void configureAllProjectsUsingConstraints(
             Project rootProject,
-            Path gradleLockfile,
+            Path _gradleLockfile,
             Map<Project, LockedConfigurations> lockedConfigurations,
             ProjectDependency locksDependency) {
 
-        List<DependencyConstraint> publishableConstraints = constructPublishableConstraintsFromLockFile(
-                gradleLockfile, rootProject.getDependencies().getConstraints());
-
-        rootProject.allprojects(subproject -> configureUsingConstraints(
-                subproject, locksDependency, publishableConstraints, lockedConfigurations.get(subproject)));
+        rootProject.allprojects(subproject ->
+                configureUsingConstraints(subproject, locksDependency, lockedConfigurations.get(subproject)));
     }
 
     private static void configureUsingConstraints(
-            Project subproject,
-            ProjectDependency locksDependency,
-            List<DependencyConstraint> publishableConstraints,
-            LockedConfigurations lockedConfigurations) {
+            Project subproject, ProjectDependency locksDependency, LockedConfigurations lockedConfigurations) {
         Configuration locksConfiguration = subproject
                 .getConfigurations()
                 .create(LOCK_CONSTRAINTS_CONFIGURATION_NAME, locksConf -> {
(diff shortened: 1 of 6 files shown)


src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -219,6 +225,23 @@ public final void apply(Project project) {
...
+        // Create "platform" configuration in root project, which will hold the strictConstraints
+        NamedDomainObjectProvider<Configuration> gcvLocksConfiguration = project.getConfigurations()
+                .register("gcvLocks", conf -> {
+                    conf.getAttributes().attribute(Usage.USAGE_ATTRIBUTE, internalUsage);
+                    conf.getOutgoing().capability(GCV_LOCKS_CAPABILITY);
+                    conf.setCanBeResolved(false);
+                    conf.setVisible(false);
+                });
+
+        ProjectDependency locksDependency =
```
ferozco, line 236: why do we create a self dependency?

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -219,6 +225,23 @@ public final void apply(Project project) {
...
+                (ProjectDependency) project.getDependencies().create(project);
+        locksDependency.capabilities(moduleDependencyCapabilitiesHandler ->
```
ferozco, line 238: is this necessary given that we already have configured the configuration?

src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
```diff
@@ -71,29 +84,39 @@ public final void apply(Project project) {
...
+
+            // Create "platform" configuration in root project, which will hold the versions props constraints
+            project.getConfigurations().register("gcvVersionsPropsConstraints", conf -> {
+                conf.getAttributes().attribute(Usage.USAGE_ATTRIBUTE, gcvVersionsPropsUsage);
+                conf.getOutgoing().capability(gcvVersionsPropsCapability);
+                conf.setCanBeResolved(false);
+                conf.setCanBeConsumed(true);
+                conf.setVisible(false);
+
+                // Note: don't add constraints to the ConstraintHandler, only call `create` / `platform` on it.
```
ferozco: What do you mean here? could you explain why?

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
dansanduleac: it's not actually "self" but pointing to a particular configuration (indirectly).
This is to pick up the locks configuration which is defined in the root project.

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
dansanduleac: yes - we will resolve multiple variants of the same project (root project) so if they don't have different (i.e. explicit and different) capabilities, they will conflict.

dansanduleac: oh right and you can't depend on an explicit capability just with attributes, you need to specify the capability.

If anything, we might not need to specify the attribute directly, but it's a bit more clear this way I think.

src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
dansanduleac: This was there before and mostly just meant the function accepting the ConstraintHandler doesn't call `add` on it, only `create`.
It's a bit confusing though so I'm gonna instead make up an interface that can only create, and pass that in.

src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
```diff
@@ -71,29 +83,38 @@ public final void apply(Project project) {
...
         VersionRecommendationsExtension extension =
                 project.getRootProject().getExtensions().getByType(VersionRecommendationsExtension.class);
 
-        VersionsProps versionsProps = loadVersionsProps(
-                project.getRootProject().file("versions.props").toPath());
-
         NamedDomainObjectProvider<Configuration> rootConfiguration = project.getConfigurations()
                 .register(ROOT_CONFIGURATION_NAME, conf -> {
                     conf.setCanBeResolved(false);
+                    conf.setCanBeConsumed(false);
```
dansanduleac, line 105: not setting this was an oversight, now fixed

carterkozak (approved):
Tested in a large internal project. Build scan is still large, but a little less so.

dansanduleac:
👍 go go go

PR #576, opened 2020-08-28
fix com.palantir.versions-props break

dansanduleac:
## Before this PR

Get breaks if certain configurations are resolved/touched during the configuration stage.
This is because in #557, we incorrectly further configure the internal configuration used to hold versions.props constraints, which sometimes breaks like this: https://github.com/palantir/gradle-revapi/pull/307

## After this PR

==COMMIT_MSG==
Fix versions-props plugin regression from #557 (1.26.0) in some circumstances 
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-576.v2.yml
+++ b/changelog/@unreleased/pr-576.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: 'Fix versions-props plugin regression from #557 (1.26.0) in some circumstances'
+  links:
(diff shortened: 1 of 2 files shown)


CRogers:
@dansanduleac Doe you think it's possible to write a test for this, or even just test it works with gradle-revapi?

dansanduleac:
tested against that gradle-revapi PR for now

CRogers:
:+1:

PR #613, opened 2020-10-28
Add property to skip writing lock file

fawind:
## Before this PR
Depending on if GCV is invoked with `--write-locks`, the unifiedClasspath might get evaluated or not: 
https://github.com/palantir/gradle-consistent-versions/blob/5d28c1cb73d2d792b750900ac669c56ec7998c3c/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java#L276-L277

In some cases this can lead to different version constraints getting added depending on if `--write-locks` is used. As a result, other plugins that use the `--write-locks` flag (e.g. [sls-packaging `createManifest` task](https://github.com/palantir/sls-packaging)) might have different results depending on if `--write-locks` is used or not.

This manifests in bugs like `./gradlew createManifest` complaining about an outdated lock file but `./gradlew createManifest --write-locks` not updating the lock file.

## After this PR

Add the optional property `gcvSkipWriteLocks` that disables writing the GCV lock file. This is intended as a workaround to unblock plugins that also use the `--write-locks` flag and run into the above problems.

E.g. you can run `./gradlew createManifest --write-locks -PgcvSkipWriteLocks` to just update the createManifest lock file without updating the GCV lock file. 

==COMMIT_MSG==
Add property to skip writing lock file
==COMMIT_MSG==

## Possible downsides?
More like a hacky workaround to unblock people being affected by the inconsistent task output. However this change should only have a small-scale impact and we can revisit with a better solution if this becomes a bigger problem.

### DIFF
--- a/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
@@ -273,12 +273,18 @@ public class VersionsLockPlugin implements Plugin<Project> {
             fullLockStateProperty.set(project.provider(fullLockStateSupplier::get));
 
             if (project.getGradle().getStartParameter().isWriteDependencyLocks()) {
-                // Triggers evaluation of unifiedClasspath
-                new ConflictSafeLockFile(rootLockfile).writeLocks(fullLockStateSupplier.get());
-                log.lifecycle("Finished writing lock state to {}", rootLockfile);
+                if (isSkipWriteLocks(project)) {
+                    log.lifecycle(
+                            "Skipped writing lock state to {} because the 'gcvSkipWriteLocks' property was set",
+                            rootLockfile);
+                } else {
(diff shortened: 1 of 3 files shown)


src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -331,6 +337,10 @@ static boolean isIgnoreLockFile(Project project) {
         return project.hasProperty("ignoreLockFile");
     }
 
+    private static boolean isSkipWriteLocks(Project project) {
+        return project.hasProperty("gcvSkipWriteLocks");
```
fawind, line 341: Happy to iterate on the prop name. Wanted to make it clear that this is specific for GCV and not other plugins that also use the `--write-locks` flag.

iamdanfox (approved):
Yep, seems like a useful debugging tool, even if I hope we don't have to use it lots!

PR #657, opened 2021-01-15
Add compatibility with Gradle configuration-on-demand

AlexLandau:
## Before this PR

Gradle configuration-on-demand is a built-in Gradle feature that is very helpful at reducing Gradle configuration times in (sufficiently-properly-configured) monorepos: https://docs.gradle.org/current/userguide/multi_project_configuration_and_execution.html#sec:configuration_on_demand

These didn't previously work well together because GCV wanted all projects to be configured during the configuration of the project (and in fact forced their configurations), while CoD configures only a subset of projects. To help deal with this incompatibility, GCV also had some direct checks telling you to disable CoD entirely in order to run key operations.

## After this PR

Now, we only force the full set of projects to be configured when using --write-locks, and otherwise allow commands like `./gradlew :some-project:build` to only configure projects that we'd expect to be configured by CoD.

This implementation is a bit of a hack, in that it makes some conservative guesses about which projects to configure rather than delegating to Gradle's "natural" configuration of those projects. I first attempted the latter, but there are too many restrictions on the timing of the various steps -- when dependencies can be defined or resolved, when we can force a project to configure itself, and so on. The good news is that the hackiness is confined to a single function, and the known inaccuracies should result in slower configuration rather than correctness errors.

## Possible downsides?

If the configuration-on-demand feature is not enabled, the code path should be effectively unchanged, making this low-risk for existing repositories.

I added tests for all the cases I could think of, but it's possible I missed something.

There is one case I didn't handle: running Gradle from somewhere other than the root directory. I believe this is uncommon, but it really depends on people's individual workflows. In this case, all the projects will be configured.

### DIFF
--- a/src/main/java/com/palantir/gradle/versions/GradleWorkarounds.java
+++ b/src/main/java/com/palantir/gradle/versions/GradleWorkarounds.java
@@ -16,12 +16,18 @@
 
 package com.palantir.gradle.versions;
 
+import groovy.lang.GString;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Proxy;
-import org.gradle.api.DomainObjectCollection;
+import java.util.LinkedHashSet;
+import java.util.Set;
 import org.gradle.api.GradleException;
+import org.gradle.api.Project;
 import org.gradle.api.ProjectState;
+import org.gradle.api.Task;
 import org.gradle.api.artifacts.Configuration;
+import org.gradle.api.artifacts.Dependency;
+import org.gradle.api.artifacts.ProjectDependency;
 import org.gradle.api.artifacts.dsl.DependencyConstraintHandler;
 import org.gradle.api.artifacts.dsl.DependencyHandler;
 import org.gradle.api.attributes.Attribute;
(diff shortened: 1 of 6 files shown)


AlexLandau:
As a side note, as part of this I wanted to update the most recent version of GradleTestVersions to 6.7.1 or 6.8, but that causes test errors due to a deprecation of something from another plugin. See: https://github.com/palantir/gradle-configuration-resolver-plugin/pull/74

src/test/groovy/com/palantir/gradle/versions/ConfigurationOnDemandSpec.groovy
```diff
@@ -0,0 +1,423 @@
...
+
+import static com.palantir.gradle.versions.GradleTestVersions.GRADLE_VERSIONS
+import static com.palantir.gradle.versions.PomUtils.makePlatformPom
+
+/**
+ * This tests the interaction of this plugin with Gradle's configuration-on-demand feature:
+ * https://docs.gradle.org/current/userguide/multi_project_configuration_and_execution.html#sec:configuration_on_demand
+ */
+@Unroll
+class ConfigurationOnDemandSpec extends IntegrationSpec {
```
CRogers, line 31: These look like great tests - very readable - thanks for writing them!

Strangely I can't run them without getting `java.lang.ClassNotFoundException: com.google.common.collect.Multimap`, but that's also true on develop with other tests, so probably a me problem.

src/main/java/com/palantir/gradle/versions/GradleWorkarounds.java
```diff
@@ -124,5 +130,82 @@ static boolean isFailOnVersionConflict(Configuration conf) {
         return conflictResolution == org.gradle.api.internal.artifacts.configurations.ConflictResolution.strict;
     }
 
+    @SuppressWarnings("CyclomaticComplexity")
+    public static void makeEvaluationDependOnSubprojectsToBeEvaluated(Project rootProject) {
+        if (!rootProject.getGradle().getStartParameter().isConfigureOnDemand()
+                || rootProject.getGradle().getStartParameter().isWriteDependencyLocks()
+                // If Gradle is run from somewhere other than the root, the task location gets trickier to translate
+                // into the projects to use; this could be implemented in the future if there's demand
+                || !rootProject.getGradle().getStartParameter().getCurrentDir().equals(rootProject.getRootDir())) {
```
CRogers, line 139: So if you run gradle from a subdir, it will still *work*, just configure on demand will not be enabled?

CRogers:
Meetings took up my afternoon/evening - I'll continue reviewing this on Monday.

src/main/java/com/palantir/gradle/versions/GradleWorkarounds.java
AlexLandau: Correct.

src/test/groovy/com/palantir/gradle/versions/ConfigurationOnDemandSpec.groovy
AlexLandau: I personally ran into some problems with "Could not load compiled classes for build file" when running tests locally, but switching the Gradle wrapper to use a non-snapshot version fixed that. (Could be a security software interaction, per [this discussion](https://discuss.gradle.org/t/gradleexception-could-not-load-compiled-classes-for-build-file-fails-on-all-scripts-even-after-removing-cache/2178/5).)

src/main/java/com/palantir/gradle/versions/GradleWorkarounds.java
carterkozak: Oh slick, I had no idea gradle supported running within a subdir to limit task breadth! Good flag.

src/main/java/com/palantir/gradle/versions/GradleWorkarounds.java
```diff
@@ -124,5 +130,82 @@ static boolean isFailOnVersionConflict(Configuration conf) {
...
+            // Just configure every project
+            rootProject.getSubprojects().forEach(subproject -> rootProject.evaluationDependsOn(subproject.getPath()));
+            return;
+        }
+
+        Set<String> projectPathsToEval = new LinkedHashSet<>();
+        for (String taskPath : rootProject.getGradle().getStartParameter().getTaskNames()) {
+            if (!taskPath.contains(":")) {
+                // This is a task to be run in each project that defines it, e.g. "build". This should cause every
+                // project to be defined in a configuration-on-demand build.
```
CRogers, line 149: Am I right in thinking we need to do this because Gradle does not automatically do it for us when in configure on demand mode? I would have thought the configuration-on-demand feature in gradle should have the behaviour of evaluating every project if you run a general task like `./gradle build`, or it just plain wouldn't work?

CRogers: Or is this just ensuring the *ordering* of project evaluation is in the correct fashion for GCV to function?

src/main/java/com/palantir/gradle/versions/GradleWorkarounds.java
```diff
@@ -124,5 +130,82 @@ static boolean isFailOnVersionConflict(Configuration conf) {
...
+
+            rootProject.evaluationDependsOn(projectPath);
+            Project project = rootProject.project(projectPath);
+            if (!project.getState().getExecuted()) {
+                throw new IllegalStateException(
+                        "The project has not yet been evaluated when we expect it to have been");
+            }
+            // Per
+            // https://docs.gradle.org/current/userguide/multi_project_configuration_and_execution.html#sec:configuration_on_demand,
+            // configuration is propagated transitively in two ways: dependencies and string-based task dependencies.
```
CRogers, line 178: If I'm reading this right, the docs seem to claim to support evaluating the projects that are used in task dependencies:

> The task dependencies declared via task path are supported and cause relevant projects to be configured. Example: someTask.dependsOn(":some-other-project:someOtherTask")

> The standard project dependencies are supported and makes relevant projects configured. If project A has a compile dependency on project B then building A causes configuration of both projects.

Yet it looks like we have similar logic in the code below? What is this doing?

CRogers:
Pretty happy with this as it is, especially as the behaviour should only be changed for projects that use configure-on-demand, so it should be pretty safe for existing users. Just have a couple of questions about the implementation so I can fully understand it - might take you up on that offer of a VTC later on if that's easier to explain it synchronously!

src/main/java/com/palantir/gradle/versions/GradleWorkarounds.java
AlexLandau: Gradle will configure those projects automatically, but not early enough in the build process for GCV's needs. Before this PR, we forced all the subprojects to be configured early with this line before proceeding:

https://github.com/palantir/gradle-consistent-versions/pull/657/files#diff-8f6332e4bf665c9817bd219407e5081faf2bd8e9b94d16b1e11d14bb2205f4e2L258

Configuring every subproject unconditionally, as that line does, would make CoD useless; but if we just remove that line with no replacement, some projects aren't configured before the point in time where we need them to know their dependencies. So we replace it with something that configures the right set of projects, or as close as we can get.

(This could perhaps work differently if writing locks happened during a task rather than during configuration, but I'm not going to try to make that large of a change, which I'm sure would cause other problems.)

src/main/java/com/palantir/gradle/versions/GradleWorkarounds.java
AlexLandau: We can't just allow Gradle to evaluate the set of projects "naturally" because it will end up too late in the build process to do all the things we want to do afterwards, so this is imitating that logic to force early evaluation of the correct set of projects.

src/main/java/com/palantir/gradle/versions/GradleWorkarounds.java
CRogers: Ah, I see, makes sense - thanks for the explanation.

CRogers:
:+1:

PR #721, opened 2021-05-20
Handle gradle 7 removing default configurations

### LINKED ISSUE #700: Gradle 7 compatibility
Is there any time frame to support gradle 7? I've tried to upgrade but looks like things break on removed legacy configurations.

https://github.com/palantir/gradle-consistent-versions/blob/ccb79b612128591efb2a829f0cb6fe6bc76585c0/src/main/java/com/palantir/gradle/versions/FixLegacyJavaConfigurationsPlugin.java#L56-L68

```
* What went wrong:
A problem occurred evaluating script.
> Failed to apply plugin class 'com.palantir.gradle.versions.FixLegacyJavaConfigurationsPlugin'.
   > Configuration with name 'compile' not found.
```

ferozco:
## Before this PR
Closes #700. Side note, its kinda funny that a gradle 7 issue is issue 700

## After this PR

==COMMIT_MSG==
Handle gradle 7 removing default configurations
==COMMIT_MSG==

## Possible downsides?
Don't think so, we are now eagerly looking for the configurations but it should be safe since we ensure that the java plugin has already been applied and hence the configurations exist

Increasing the minimum required Gradle version from 5.x to 6.1.


carterkozak:
Can we add a gradle 7 integration test?

src/main/java/com/palantir/gradle/versions/GradleConfigurations.java
```diff
@@ -16,14 +16,23 @@
...
 import org.gradle.api.plugins.JavaPluginConvention;
+import org.gradle.api.tasks.SourceSet;
 
 final class GradleConfigurations {
+    /**
+     * Deprecated sourcesets of the compile and runtime configuration, to be removed with Gradle 7.
+     * The full configuration name follows the naming scheme of "$taskBaseName + capitalize($configurationName)"
+     * see {@link org.gradle.api.internal.tasks.DefaultSourceSet#configurationNameOf}.
+     */
+    private static final ImmutableList<String> DEPRECATED_SOURCESET_SUFFIXES = ImmutableList.of("Compile", "Runtime");
```
fawind, line 34: The `getCompileConfigurationName` and `getRuntimeConfigurationName` method no longer exist with Gradle 7 and the method for building the configuration name based on the base config is private ([ref](https://github.com/gradle/gradle/blob/a2ea10896cbdca85a4fb7b0e30d638f57dbe1533/subprojects/plugins/src/main/java/org/gradle/api/internal/tasks/DefaultSourceSet.java#L143-L145)).

Matching on the configuration name suffix for backwards compatibility but for Gradle 7 projects, this shouldn't match anything and can be removed in the future.

src/test/groovy/com/palantir/gradle/versions/ConsistentVersionsPluginIntegrationSpec.groovy
```diff
@@ -41,24 +41,32 @@ class ConsistentVersionsPluginIntegrationSpec extends IntegrationSpec {
...
-                dependencies {
-                    classpath 'com.palantir.configurationresolver:gradle-configuration-resolver-plugin:0.3.0'
-                }
-            }
             plugins {
                 id '${PLUGIN_NAME}'
             }
             allprojects {
-                apply plugin: 'com.palantir.configuration-resolver'
+                tasks.register("resolveConfigurations", {
```
fawind, line 48: I haven't looked too deeply into why, but on Gradle 7, we fail to resolve the `gradle-configuration-resolver-plugin`. However given that the [plugin is archived](https://github.com/palantir/gradle-configuration-resolver-plugin), I pulled the relevant task out into directly into this test and removed the dependency.

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -103,7 +103,7 @@
 
 public class VersionsLockPlugin implements Plugin<Project> {
     private static final Logger log = Logging.getLogger(VersionsLockPlugin.class);
-    static final GradleVersion MINIMUM_GRADLE_VERSION = GradleVersion.version("5.3");
+    static final GradleVersion MINIMUM_GRADLE_VERSION = GradleVersion.version("6.1");
```
fawind, line 106: This is imposed by `nebula-dependency-recommender:10.0.0`.

src/main/java/com/palantir/gradle/versions/FixLegacyJavaConfigurationsPlugin.java
```diff
@@ -54,17 +55,16 @@ public final void apply(Project project) {
     }
 
     private void fixLegacyResolvableJavaConfigurations(Project project, Configuration unifiedClasspath) {
-        Stream.of(
-                        JavaPlugin.COMPILE_CONFIGURATION_NAME,
-                        JavaPlugin.COMPILE_ONLY_CONFIGURATION_NAME,
-                        JavaPlugin.RUNTIME_CONFIGURATION_NAME)
-                .map(project.getConfigurations()::named)
-                .forEach(confProvider -> confProvider.configure(conf -> {
+        // TODO(fwindheuser): Remove compile and runtime after stating to build with Gradle 7+
```
carterkozak, line 58: Perhaps we should enforce this by adding a precondition that `MINIMUM_GRADLE_VERSION` < 7?

fawind: I think this precondition would get triggered when a consuming repository builds with Gradle 7? I think for now, we want to enable users to use either Gradle 6 or 7.

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
CRogers: We only need nebula-dependency-recommender for tests, right? In theory, we could replace it with our own gradle code that does the same stuff.

That said, I think it's fine to kill off Gradle 5 support now that 7 is out (probably want to major rev it).

CRogers (approved):

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
fawind: Yes, it's just for tests so not strictly raising the minimum gradle version for the actual plugin.

PR #723, opened 2021-06-16
Write locks task to enable writing just GCV dependency locks

CRogers:
## Before this PR
Currently, you need to run `--write-locks` in order to write out the `versions.lock` file. However, a lot of other infra has _also_ latched onto the `--write-locks`, meaning when devs want to update just their `versions.lock`, they have no choice but to update everything, which can take a very long time. This is only going to get worse with the Per Endpoint Minimum Version work, which requires compilation upon `--write-locks`, making it take significantly longer.

## After this PR

==COMMIT_MSG==
A new `writeVersionsLock` task that allows writing out the `versions.lock` without running `--write-locks` and possibly triggering expensive code paths from other gradle plugins.
==COMMIT_MSG==

## Possible downsides?
Have to use a funky "marker task" in order to maintain the behaviour of writing locks out at `project.afterEvaluate` time rather than task execution time. This means that something like `myTask.dependsOn writeVersionsLock` will not write the locks if you do `./gradle myTask`, as it only looks for the `writeVersionsLock` in the gradle start parameter list.

Should probably think carefully about what to name this task, as people we write it often (although can be abbreviated in the standard gradle way - `./gradlew wVL`)


src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -120,6 +121,7 @@
     private static final Attribute<GcvUsage> GCV_USAGE_ATTRIBUTE =
             Attribute.of("com.palantir.consistent-versions.usage", GcvUsage.class);
     private static final String GCV_LOCKS_CAPABILITY = "gcv:locks:0";
+    private static final String WRITE_VERSIONS_LOCK_TASK = "writeVersionsLock";
```
CRogers: Commence bikeshedding here.

Here are a few options and my thoughts:

1. `writeVersionsLock` - communicates clearly we're updating the `versions.lock` file, but is a bit of a mouthfull
2. `versionsLock` - a bit shorter than 1) but doesn't imply we're updating it
3. `gcvLocks` - if you know what GCV is and how it works, this is obvious, but maybe not as obvious as 1)
4. `writeDependencyLocks` - probably the most similar to the current `--write-locks`, although the use of "dependency" may not be as clear.

carterkozak: I like the idea of a `writeLocks` prefix or suffix to group write-locks tasks together. Neither `writeLocksVersions` nor `versionsWriteLocks` exactly roll off the tongue though.

carterkozak: So I suppose my preference is the sandwich pattern `write[identifier]Lock(s)?` Not sure if we want lock to be singular or plural, "write version locks" vs "write versions lock".

I think plural form "Locks" may be slightly better because some lock tasks update multiple files, e.g. `[module]/baseline-class-uniqueness.lock`, and we probably want the names to be as similar as possible to reduce mental overhead.

I don't have a terribly strong opinion here, just typing out my thoughts.

esword: +1 to writeVersionsLocks

fawind (approved):
Change looks good!

Also don't have a strong opinion on the name of the task, I think `writeVersionsLock(s)` is good but will let you decide :)

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -280,29 +288,33 @@ public final void apply(Project project) {
             });
             fullLockStateProperty.set(project.provider(fullLockStateSupplier::get));
 
-            if (project.getGradle().getStartParameter().isWriteDependencyLocks()) {
-                if (isSkipWriteLocks(project)) {
-                    log.lifecycle(
-                            "Skipped writing lock state to {} because the 'gcvSkipWriteLocks' property was set",
-                            rootLockfile);
+            project.getGradle().getTaskGraph().whenReady(taskExecutionGraph -> {
```
CRogers: Ok, I've made a little change here which _I think_ should be safe (but as always with Gradle...). We wait for the task graph to be evaluated before writing the lock file, so that abbreviating task names still works (like `wVL`, `wVersionLock`) which should make dev's lives a little easier.

CRogers: Unsurprisingly this does not work. I think task graph evaluation time is too late.

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
CRogers: I've gone with `writeVersionsLocks` as with gradle-like task abbreviation all of these work:

```
writeVersionsLocks
writeVersionsLock
writeVersionLocks
wVL
```

carterkozak (approved):
lgtm. Possible that something like `wVL` will become API, but not terribly likely, nor is it a big deal.

src/main/java/com/palantir/gradle/versions/TaskNameMatcher.java
```diff
@@ -0,0 +1,65 @@
...
+        if (taskNameParts.size() != fullTaskNameParts.size()) {
+            return false;
+        }
+
+        return Streams.zip(fullTaskNameParts.stream(), taskNameParts.stream(), String::startsWith)
+                .allMatch(bool -> bool);
+    }
+
+    public boolean matchesAny(Collection<String> tasks) {
+        return tasks.stream().allMatch(this::matches);
```
fawind: Shouldn't this be `.anyMatch()` or am I reading it wrong?

CRogers: You are indeed correct!

CRogers: Me a couple of hours ago:

> This is super simple, I don't need to test this

carterkozak: Me a couple minutes ago:

> There's a test case, I can skim the logic quickly

:man_facepalming:

src/main/java/com/palantir/gradle/versions/TaskNameMatcher.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+final class TaskNameMatcher {
+    private final List<String> fullTaskNameParts;
+
+    TaskNameMatcher(String fullTaskName) {
+        List<String> parts = toParts(fullTaskName);
+
+        this.fullTaskNameParts = parts.stream().filter(part -> !part.isEmpty()).collect(Collectors.toList());
```
fawind: I guess it doesn't hurt to double check but we already filter out empty parts in the `toParts` method.

CRogers:
> Possible that something like wVL will become API, but not terribly likely, nor is it a big deal.

True, but this is currently a problem with any gradle task. I actually wish you could enable gradle to not do task abbreviation when not in an interactive terminal, as using abbreviations in scripts is just waiting to explode.

src/main/java/com/palantir/gradle/versions/TaskNameMatcher.java
CRogers: Good spot, had moved this to the `toParts` method and forgot to remove it.

PR #724, opened 2021-06-16
add versions.list as clearer versions.lock diff target

dwyand:
## Before this PR

The versions.lock file is great for gradle and for digging into dependency trees with hashes, but it creates very noisy diffs when small dependency changes affect many constraint counts and hashes, making it difficult to hunt down actual dependency changes.

## After this PR
==COMMIT_MSG==
Add a versions.list file alongside the versions.lock that excludes constraint counts and hashes, providing concise human-readable diffs.
==COMMIT_MSG==

## Possible downsides?
This produces two files with the same information, formatted differently. Since they are automatically generated, they should remain in sync, but it could be confusing if they diverge.

## Bikeshedding Preemption
* The new file is named versions.list specifically to come before versions.lock in diff UIs that order files alphabetically (in particular, on github PR views).
* The new file is generated by default, which may surprise users on upgrade; but I felt it was better to advertise the feature than leave something hidden away behind configuration described in documentation that nobody reads.
* While there is a parameter to disable it, similar to `gcvSkipWriteLocks`, this may result in a weird state where `versions.lock` is inconsistent with `versions.list`. If this happens, and a `versions.list` file exists, gradle will emit a warning.


iamdanfox:
Defer to current GCV maintainers for the official review, but I personally don't think this extra file should be imposed on every single repo using GCV, as I'd suggest the majority are actually small enough that changes to versions.lock remain quite readable.  For example, see PRs like this: https://github.com/palantir/dialogue/pull/1273.

Could I suggest that if you just want a checked-in listing of all the maven coordinates for a specific configuration for git diff purposes, that you just use Gradle's built-in lockfiles?? https://docs.gradle.org/current/userguide/dependency_locking.html 

In your root build.gradle, you could add something like:

```gradle
configurations {
    unifiedClasspath {
        resolutionStrategy.activateDependencyLocking()
    }
}
```

If you want to change the filename/location you can do that too https://docs.gradle.org/current/userguide/dependency_locking.html#configuring_the_per_project_lock_file_name_and_location

PR #725, opened 2021-06-16
Delay configuration-on-demand-related errors until task execution

AlexLandau:
This was causing an issue in a situation in which the task was
getting instantiated but not actually run.

## Before this PR
Using certain internal plugins could cause the `:checkUnusedConstraints` to be instantiated and this failure to happen, although it would be filtered out of the set of tests to run.

## After this PR
==COMMIT_MSG==
Fix the `checkUnusedConstraints` task to delay throwing an error about incomplete configuration due to configure-on-demand until the task is actually run. This fixes some situations in which the task is instantiated but not run.
==COMMIT_MSG==

This also makes it play nicely with `--continue`.

## Possible downsides?

None that I'm aware of.

### DIFF
--- a/src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsTask.java
+++ b/src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsTask.java
@@ -30,6 +30,7 @@ import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.gradle.api.DefaultTask;
+import org.gradle.api.GradleException;
 import org.gradle.api.Project;
 import org.gradle.api.artifacts.component.ModuleComponentIdentifier;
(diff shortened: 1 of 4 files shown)


src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsTask.java
```diff
@@ -72,6 +76,15 @@ final void setPropsFile(File propsFile) {
         return propsFileProperty;
     }
 
+    @Input
+    public final Property<Boolean> getShouldFailWithConfigurationOnDemandMessage() {
+        return shouldFailWithConfigurationOnDemandMessage;
+    }
+
+    final void setShouldFailWithConfigurationOnDemandMessage(boolean shouldFail) {
```
CRogers, line 84: I don't think this overload is necessary - the property can just be used.

CRogers (approved):

PR #799, opened 2021-11-15
Add plugin and task for checking for new versions

markelliot:
Pushing up this PR for discussion before getting deeper into functionality or tests.

## Before this PR
gcv enables setting of and locking versions per props and config files, but there's no
facility for detecting when to apply upgrades. While not strictly required for gcv, it
would be convenient in the context of managing versions to report what upgrades may be
applied.

## After this PR

==COMMIT_MSG==
Add task for checking for new versions of dependencies
==COMMIT_MSG==

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -90,6 +90,14 @@ gradlePlugin {
             id = 'com.palantir.versions-props'
             implementationClass = 'com.palantir.gradle.versions.VersionsPropsPlugin'
         }
+        getVersions {
+            id = 'com.palantir.get-versions'
+            implementationClass = 'com.palantir.gradle.versions.GetVersionPlugin'
+        }
+        upgradeVersions {
+            id = 'com.palantir.check-versions'
+            implementationClass = 'com.palantir.gradle.versions.CheckVersionsPlugin'
+        }
     }
 }
 
--- a/src/main/java/com/palantir/gradle/versions/CheckNewVersionsTask.java
+++ b/src/main/java/com/palantir/gradle/versions/CheckNewVersionsTask.java
@@ -17,7 +17,6 @@
 package com.palantir.gradle.versions;
 
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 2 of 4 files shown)


palantirtech:
Thanks for your interest in [palantir/gradle-consistent-versions](https://github.com/palantir/gradle-consistent-versions), @markelliot! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

markelliot:
When the tests run a report like the following is generated:
```
> Task :foo:checkNewVersions
Dependency upgrades for project 'foo' configuration 'compileClasspath':
 - org.slf4j:slf4j-api 1.7.24 -> 1.7.25
Dependency upgrades for project 'foo' configuration 'default':
 - org.slf4j:slf4j-api 1.7.24 -> 1.7.25
Dependency upgrades for project 'foo' configuration 'implementation':
 - org.slf4j:slf4j-api 1.7.24 -> 1.7.25
Dependency upgrades for project 'foo' configuration 'runtimeClasspath':
 - org.slf4j:slf4j-api 1.7.24 -> 1.7.25
Dependency upgrades for project 'foo' configuration 'runtimeElements':
 - org.slf4j:slf4j-api 1.7.24 -> 1.7.25
Dependency upgrades for project 'foo' configuration 'testCompileClasspath':
 - org.slf4j:slf4j-api 1.7.24 -> 1.7.25
Dependency upgrades for project 'foo' configuration 'testImplementation':
 - org.slf4j:slf4j-api 1.7.24 -> 1.7.25
Dependency upgrades for project 'foo' configuration 'testRuntimeClasspath':
 - org.slf4j:slf4j-api 1.7.24 -> 1.7.25
```

Some discussion items:
1. Report format:
   - we could consolidate the report so that it doesn't list by configuration
   - we could group configurations with the same recommendation together
2. Enforcement:
   - we could add an 'enforce' mode that makes the task fail when dependencies have not been upgraded to the latest version
3. ResolutionStrategy
    - I haven't found a way to ignore ResolutionStrategy forced version overrides, but also I'm not sure if it's practical to override these versions as some recommendations coerce one dependency to another
4. Relevance:
    - It's not strictly necessary to bundle this plugin to get this functionality, if it's not desired in this project I'll isolate it, but chatting with @ferozco it sounded like it might be generally relevant for users of this project
    - It might be helpful to have an 'apply' or 'interactive' mode for the new versions to update the .props file produced by this plugin bundle, it's more practical to do that when it's safe to assume gcv is in use, which suggests colocating this task is a good idea

fawind:
Thanks for the PR! To be honest, I don't think we will get much use out of it internally, as the workflow of updating dependencies is pretty much covered by excavator. However, I can see it being handy for other external consumers of the library and think it would be a nice addition.

To address some of your points:

> 1. Report format:
we could consolidate the report so that it doesn't list by configuration
we could group configurations with the same recommendation together


If we bundle this with GCV, I think it would be nice to consolidate the report based on dependencies. Because when using GCV, having an outdated version in your versions.props will result in all projects and all configurations that use that dep getting flagged as out-of-date. Not sure how helpful this is to users as all these warnings have the same root cause (an old version in versions.props).

> 3. ResolutionStrategy
I haven't found a way to ignore ResolutionStrategy forced version overrides, but also I'm not sure if it's practical to override these versions as some recommendations coerce one dependency to another

Without having dived into the code too much, what is the current behavior? If I have a version override of an outdated version, will this get flagged?

markelliot:
Thanks for taking the time to take a look at this!

w/r/t relevance I think you're right that this functionality is on the edge of gcv's scope -- my original suspicion was that the versions.props/versions.lock code was going to make a big impact on if/how to assess new valid versions and if/how it'd be easy to make updates to versions.props files. Having gotten into things a little more it seems like figuring out new versions can happen completely independently of gcv, and that updating versions.props file programmatically will need some utility that differs from what gcv would normally consume, so I'm going to pull the whole set of functionality to an independent (but compatible) plugin https://github.com/markelliot/gradle-versions.

Sorry if this wasted some of your time =/

PR #815, opened 2021-12-01
Provide constraints for java libraries in the same repo

carterkozak:
==COMMIT_MSG==
Provide constraints for java libraries in the same repo
==COMMIT_MSG==

PR #821, opened 2021-12-06
Revert "Provide constraints for java libraries in the same repo (#815)"

fawind:
## Before this PR
In https://github.com/palantir/gradle-consistent-versions/pull/815, we started adding constraints for published java-libraries in the same repo.

Multiple projects are failing to resolve transitive project dependencies after picking up these changes. The two scenarios encountered so fat:
* Projects producing a bundle in which they explicitly want to depend on a previous publish of a sub-project dependency.
* Projects that pull in a sub-project dependency transitively.

In both cases, Gradle adds a constraint on the sub-project dependency, bumping the version to the local project version. However it then fails to resolve the sub-project dependency as it tries to look-up the local project version in external repositories.

*Edit:* I looked a bit more into the failures and I think the issues we saw are all related to projects having cyclic dependencies. E.g. in one example we had a cyclic transitive dependency of `project -> external-dep -> project`. Previously, this resolved fine because the transitive `project` dep could be resolved using a previously published version. However now that we added local project constraints, the version of the transitive `project` dep gets bumped to the local project version which has not been build yet and thus cannot be resolved.

## After this PR

Reverting for now. We still want the overall outcome but have to figure out a solution for these edge cases.

==COMMIT_MSG==
Revert "Provide constraints for java libraries in the same repo (#815)"
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-821.v2.yml
+++ b/changelog/@unreleased/pr-821.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Revert "Provide constraints for java libraries in the same repo (#815)"
+  links:
+  - https://github.com/palantir/gradle-consistent-versions/pull/821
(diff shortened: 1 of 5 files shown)


carterkozak (approved):
Thanks!

PR #825, opened 2021-12-09
Provide constraints for java libraries in the same repo, behind experimental feature flag

carterkozak:
## Before this PR
Publications from a single repo could be resolved to different versions -- e.g. higher version API than implementation.
The previous attempt to merge this resulted in issues handling circular dependencies, we we're hiding the feature behind a flag until we've been able to resolve the edge cases.

## After this PR
==COMMIT_MSG==
Provide constraints for java libraries in the same repo, behind experimental feature flag: `com.palantir.gradle.versions.publishLocalConstraints`
==COMMIT_MSG==


fawind (approved):
Looks good!

PR #871, opened 2022-02-23
Revert Annotation Processor Inclusion

carterkozak:
This fixes an issue in which error-prone annotations would transitively cause dependent projects to use newer-than-expected error-prone releases.
In general locking annotation processors is desirable, however we must find a way to decouple compiler plugins from annotation processors before we can roll this out successfully.

==COMMIT_MSG==
Annotation processors are no longer included in locked configurations by default
==COMMIT_MSG==

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,8 +136,8 @@ com.github.zafarkhaja:java-semver:0.9.0 (1 constraints: c315c0d2)
 com.jayway.awaitility:awaitility:1.6.5 (1 constraints: c615c1d2)
 ```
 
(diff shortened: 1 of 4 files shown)


fawind:
See https://github.com/palantir/gradle-baseline/pull/2069 for more info on the error-prone issues we are running into (i.e. this is blocking the error-prone version used in gradle-baseline to 2.10.0 while other repos already use 2.11.0).

pkoenig10 (approved):

fawind (approved):
Sad but agree that the error-prone issues are bigger than the few compile issues we saw due to different immutable versions being used (which motivated this feature).

PR #884, opened 2022-03-10
Reduce noisy logs from info to debug

pkoenig10:
I frequently use `--info` to debug Gradle build issues. These two log lines are quite noisy and produce a ton of output. They don't seem particularly relevant in most cases, so this PR lowers the level to debug.


src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -613,7 +613,7 @@ private void recursivelyCopyProjectDependenciesWithScope(
 
             Configuration targetConf = getTargetConfiguration(dependencySet, projectDependency);
 
-            log.info(
+            log.debug(
```
carterkozak: maybe isDebugEnabled to avoid `formatProjectDependency`?

carterkozak (approved):

PR #916, opened 2022-04-15
Allow passing multiple hashes to the `why` task

AlexLandau:
Running this task multiple times can be somewhat slow in large
repositories. Allow passing in multiple, comma-delimited hashes
to the `--hash` argument to receive information about multiple
dependencies in a single run.

## Before this PR

There's no way to ask the `why` task about multiple dependencies in a single Gradle run -- you'd have to ask in consecutive runs, which incurs the cost of Gradle starting a run and gathering information.

## After this PR

==COMMIT_MSG==
The `--hash` argument of the `why` task now accepts multiple, comma-separated hashes.
==COMMIT_MSG==

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,9 @@ This is effectively just a more concise version of `dependencyInsight`:
 ./gradlew  dependencyInsight --configuration unifiedClasspath --dependency jackson-databind
 ```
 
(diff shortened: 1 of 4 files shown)


carterkozak (approved):
Love it, thanks!

PR #948, opened 2022-05-23
only load versions.props once

rzpt:
## Before this PR

versions.props was loaded and parsed once per project.

## After this PR

==COMMIT_MSG==
versions.props is only loaded once
==COMMIT_MSG==

## Possible downsides?

I'm not sure what the ideal way to write this is.


fawind (approved):
Makes sense!

PR #968, opened 2022-08-03
Support lenient versions in GCV to enable Jakarta bridge migrations

mglazer:
## Before this PR

We were unable to run subprojects with different versions from the `versions.lock`

## After this PR

==COMMIT_MSG==
Support lenient versions in GCV to enable Jakarta bridge migrations

In order to support libraries like tracing-java which need
both javax libraries and jakarta libraries, we need to have some
way of hosting both in the same repository, but not having GCV
force our versions of all subprojects up to that version as well.

This introduces a mode in GCV (which is very much frowned upon
anyone using), which instead of forcing strict versions from the
versions.lock file, instead makes them prefer, this then lets
subprojects force the version to strict, which overrides any
past or future version constraints.

I'm not super happy with this solution since it does violate some
of the principles of GCV, but the problem we're trying to solve with
the Jakarta migration is very much non-compliant.
==COMMIT_MSG==

## Possible downsides?

Anyone but a small handful of projects uses this new feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,19 +47,20 @@ Direct dependencies are specified in a top level `versions.props` file and then 
     1. An evolution of `nebula.dependency-recommender`
 1. [Concepts](#concepts)
     1. versions.props: lower bounds for dependencies
-    1. versions.lock: compact representation of your prod classpath
-    1. ./gradlew why
-    1. ./gradlew checkUnusedConstraints
-    1. getVersion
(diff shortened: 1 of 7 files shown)


mglazer:
Unfortunately testing this out on a real project in integration doesn't work as I'd have hoped, the strict locks from the subprojects prevent the configuration from being resolvable.

PR #1049, opened 2023-05-16
Ensure overbroad wildcard constraints are not used in the versions.props

rzpt:
## Before this PR
Overly broad version pins cause slow gradle resolution times.

## After this PR
==COMMIT_MSG==
Add checkBadPins task that verifies that there are no overly broad version pins.
==COMMIT_MSG==

## Possible downsides?
The version pins the task adds are not ideal.


src/main/java/com/palantir/gradle/versions/CheckBadPinsTask.java
```diff
@@ -0,0 +1,166 @@
...
+                    .lifecycle("Adding pins to versions.props:\n"
+                            + newLines.stream().collect(Collectors.joining("\n")));
+            writeVersionsProps(getPropsFile().get().getAsFile(), newLines);
+            return;
+        }
+
+        throw new RuntimeException("There are efficient pins missing from your versions.props: \n"
+                + newLines
+                + "\n\n"
+                + "Run ./gradlew checkBadPins --fix to remove them.");
```
FinlayRJW: `Run ./gradlew checkBadPins --fix to add them.`

src/main/java/com/palantir/gradle/versions/CheckBadPinsTask.java
```diff
@@ -0,0 +1,166 @@
...
+        Map<String, Line> lineByArtifact = lockState.allLines().stream()
+                .collect(Collectors.toMap(line -> line.identifier().toString(), line -> line));
+        Set<String> artifacts = lineByArtifact.keySet();
+
+        Set<String> exactConstraints = versionsProps.getFuzzyResolver().exactMatches();
+        Set<String> unmatchedArtifacts = new HashSet<>(Sets.difference(artifacts, exactConstraints));
+
+        List<String> newLines = new ArrayList<>();
+        // assumes globs are sorted by specificity
+        for (FuzzyPatternResolver.Glob glob : versionsProps.getFuzzyResolver().globs()) {
```
FinlayRJW: nit: could use a steam here

```java
versionsProps.getFuzzyResolver().globs().forEach(glob -> {
            Set<String> matchedByGlob =
                    unmatchedArtifacts.stream().filter(glob::matches).collect(Collectors.toSet());
            unmatchedArtifacts.removeAll(matchedByGlob);
            newLines.addAll(computeNewLines(getVersionPin(versionsProps, glob), matchedByGlob, lineByArtifact));
        });
```

src/main/java/com/palantir/gradle/versions/CheckBadPinsTask.java
```diff
@@ -0,0 +1,166 @@
...
+import org.gradle.api.file.RegularFile;
+import org.gradle.api.file.RegularFileProperty;
+import org.gradle.api.provider.Property;
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.InputFile;
+import org.gradle.api.tasks.TaskAction;
+import org.gradle.api.tasks.options.Option;
+import org.gradle.language.base.plugins.LifecycleBasePlugin;
+
+public class CheckBadPinsTask extends DefaultTask {
```
CRogers: abstract class

src/main/java/com/palantir/gradle/versions/CheckBadPinsTask.java
```diff
@@ -0,0 +1,166 @@
...
+
+    private final Property<Boolean> shouldFix = getProject().getObjects().property(Boolean.class);
+    private final RegularFileProperty propsFileProperty =
+            getProject().getObjects().fileProperty();
+    private final RegularFileProperty lockFileProperty =
+            getProject().getObjects().fileProperty();
```
CRogers: abstract properties

src/main/java/com/palantir/gradle/versions/CheckBadPinsTask.java
CRogers: I'd call this `CheckOverbroadConstraints`

src/test/groovy/com/palantir/gradle/versions/CheckBadPinsIntegrationSpec.groovy
```diff
@@ -0,0 +1,95 @@
...
+        """.stripIndent().trim()
+
+        then:
+        buildAndFailWith('There are efficient pins missing from your versions.props: \n[com.fasterxml.jackson.core:jackson-annotations = 2.9.5]')
+        buildWithFixWorks()
+        file('versions.props').text.trim() == """
+            com.fasterxml.jackson.*:* = 2.9.3
+            com.fasterxml.jackson.core:jackson-annotations = 2.9.5
+        """.stripIndent().trim()
+    }
```
CRogers: Needs a test to show how fix works.

src/main/java/com/palantir/gradle/versions/CheckBadPinsTask.java
```diff
@@ -0,0 +1,166 @@
...
+    public final void checkBadPins() {
+        VersionsProps versionsProps =
+                VersionsProps.loadFromFile(getPropsFile().get().getAsFile().toPath());
+        LockState lockState =
+                new ConflictSafeLockFile(getLockFile().get().getAsFile().toPath()).readLocks();
+
+        checkBadPins(versionsProps, lockState);
+    }
+
+    private void checkBadPins(VersionsProps versionsProps, LockState lockState) {
```
CRogers: Might be possible to unit test just this method (rather than running Gradle-based tests for all of them), so we can easily list all the possible examples.

src/main/java/com/palantir/gradle/versions/CheckBadPinsTask.java
```diff
@@ -0,0 +1,166 @@
...
+                    unmatchedArtifacts.stream().filter(glob::matches).collect(Collectors.toSet());
+            unmatchedArtifacts.removeAll(matchedByGlob);
+            newLines.addAll(computeNewLines(getVersionPin(versionsProps, glob), matchedByGlob, lineByArtifact));
+        }
+
+        if (newLines.isEmpty()) {
+            return;
+        } else if (shouldFix.get()) {
+            getProject()
+                    .getLogger()
```
CRogers: `getProject()` in task actions is illegal, use a `Logging.getLogger` instead.

src/main/java/com/palantir/gradle/versions/CheckBadPinsTask.java
```diff
@@ -0,0 +1,166 @@
...
+        getOutputs().upToDateWhen(_task -> true); // task has no outputs, this is needed for it to be up to date
+    }
+
+    final void setLockFile(File lockFile) {
+        this.lockFileProperty.set(lockFile);
+    }
+
+    final void setPropsFile(File propsFile) {
+        this.propsFileProperty.set(propsFile);
+    }
```
CRogers: These are not needed when abstract.

src/main/java/com/palantir/gradle/versions/CheckBadPinsTask.java
```diff
@@ -0,0 +1,166 @@
...
+
+    @InputFile
+    public final Property<RegularFile> getLockFile() {
+        return lockFileProperty;
+    }
+
+    @InputFile
```
CRogers: When combined in the abstract version, this should be `@OutputFile` too

src/test/groovy/com/palantir/gradle/versions/CheckBadPinsIntegrationSpec.groovy
```diff
@@ -0,0 +1,95 @@
...
+
+class CheckBadPinsIntegrationSpec extends IntegrationSpec {
+
+    def setup() {
+        buildFile << """
+            plugins {
+                id 'com.palantir.versions-lock'
+                id 'com.palantir.versions-props'
+            }
+        """.stripIndent()
```
CRogers: `.stripIndent(true)`

CRogers: Many times in this file.

src/main/java/com/palantir/gradle/versions/CheckBadPinsTask.java
```diff
@@ -0,0 +1,166 @@
...
+                .map(line -> String.format("%s = %s", line.identifier(), line.version()))
+                .collect(Collectors.toList());
+    }
+
+    private static void writeVersionsProps(File propsFile, List<String> newLines) {
+        List<String> lines = readVersionsPropsLines(propsFile);
+        try (BufferedWriter writer0 =
+                        Files.newBufferedWriter(propsFile.toPath(), StandardOpenOption.TRUNCATE_EXISTING);
+                PrintWriter writer = new PrintWriter(writer0)) {
+            Stream.of(lines, newLines).flatMap(List::stream).forEach(writer::println);
```
CRogers: This is a complicated way (but slightly more efficient way) of doing this. Since we know the file is small, I'd just build a string in memory using `.collect(Collectors.joining("\n")` and write it out with `Files.writeString`.

src/main/java/com/palantir/gradle/versions/CheckBadPinsTask.java
CRogers: Actually, I don't think gradle likes this, the `outputs` thing in the constructor is the correct thing.

src/main/java/com/palantir/gradle/versions/CheckBadPinsTask.java
FinlayRJW: had to annotate with `@Internal` is that expected

src/main/java/com/palantir/gradle/versions/CheckBadPinsTask.java
FinlayRJW: I also updated the name to be `checkOverbroadConstraints` for running from gradle i.e. `./gradlew checkOverbroadConstraints`

src/test/groovy/com/palantir/gradle/versions/CheckBadPinsIntegrationSpec.groovy
FinlayRJW: test already covers fix - have renamed test to make it clearer

src/main/java/com/palantir/gradle/versions/CheckBadPinsTask.java
FinlayRJW: I'm not quite sure of the value of this over just using the integration spec - I have pulled out a static method `determineNewLines` which we can now test independently but I can only think of two relevant tests, versions props has all pins or versions props is missing pins - maybe there is something I've missed here but feel my tests just duplicate what is already being done

src/main/java/com/palantir/gradle/versions/CheckOverbroadConstraints.java
```diff
@@ -0,0 +1,165 @@
...
+
+    @InputFile
+    public final Property<RegularFile> getPropsFile() {
+        return getPropsFileProperty();
+    }
+
+    @Input
+    public final Property<Boolean> getShouldFix() {
+        return getShouldFixProperty();
+    }
```
CRogers: We don't need these at all, just name the property methods without `Property` at the end.

src/main/java/com/palantir/gradle/versions/CheckOverbroadConstraints.java
```diff
@@ -0,0 +1,165 @@
...
+            return;
+        } else if (getShouldFixProperty().get()) {
+            getLogger()
+                    .lifecycle("Adding pins to versions.props:\n"
+                            + newLines.stream().collect(Collectors.joining("\n")));
+            writeVersionsProps(getPropsFile().get().getAsFile(), newLines);
+            return;
+        }
+
+        throw new RuntimeException("There are efficient pins missing from your versions.props: \n"
```
CRogers: Need a better exception message than this. Maybe we need a good one line description of the problem and a readme section we can link to. There's a good explanation of the problem [here](https://pl.ntr/2oW) that we could probably take.

src/main/java/com/palantir/gradle/versions/CheckOverbroadConstraints.java
CRogers: And move the `@InputFile` etc with them instead of `@Internal`

src/main/java/com/palantir/gradle/versions/CheckBadPinsTask.java
CRogers: It's only worth it if there are a lot of tests and we don't want to run the version that runs an entire Gradle every time.

src/main/java/com/palantir/gradle/versions/CheckOverbroadConstraints.java
FinlayRJW: Hmm, I have configured these such that 

```java
    @Input
    public abstract Property<Boolean> getShouldFix();
```

but I am getting `Cannot have abstract method CheckOverbroadConstraints.getShouldFix()` - what am I missing here?

src/main/java/com/palantir/gradle/versions/CheckBadPinsTask.java
FinlayRJW: I'll leave both as I can see this getting enhancements in the future where more tests are required

src/main/java/com/palantir/gradle/versions/CheckOverbroadConstraints.java
FinlayRJW: I've added to the readme and improve the exception message

src/main/java/com/palantir/gradle/versions/CheckOverbroadConstraints.java
```diff
@@ -97,18 +80,22 @@ private void checkOverbroadConstraints(VersionsProps versionsProps, LockState lo
...
+        } else if (getShouldFix().get()) {
             getLogger()
                     .lifecycle("Adding pins to versions.props:\n"
                             + newLines.stream().collect(Collectors.joining("\n")));
             writeVersionsProps(getPropsFile().get().getAsFile(), newLines);
             return;
         }
 
-        throw new RuntimeException("There are efficient pins missing from your versions.props: \n"
+        throw new RuntimeException("Over-broad version constraints found in versions.props.\n"
```
CRogers: Use `String.join("\n` usually makes this easier to read.

src/main/java/com/palantir/gradle/versions/CheckOverbroadConstraints.java
```diff
@@ -97,18 +80,22 @@ private void checkOverbroadConstraints(VersionsProps versionsProps, LockState lo
...
+                + "Over-broad constrains often arise due to wildcards in versions.props\n"
+                + "which apply to more dependencies than they should, this can lead to slow builds.\n"
+                + "The following additional pins are recommended:\n"
                 + newLines
                 + "\n\n"
-                + "Run ./gradlew checkOverbroadConstraints --fix to add them.");
+                + "Run ./gradlew checkOverbroadConstraints --fix to add them."
+                + "See https://pl.ntr/2oX for details");
```
CRogers: Link to the public readme not the internal thing

src/main/java/com/palantir/gradle/versions/CheckOverbroadConstraints.java
```diff
@@ -97,18 +80,22 @@ private void checkOverbroadConstraints(VersionsProps versionsProps, LockState lo
...
 
-        throw new RuntimeException("There are efficient pins missing from your versions.props: \n"
+        throw new RuntimeException("Over-broad version constraints found in versions.props.\n"
+                + "Over-broad constrains often arise due to wildcards in versions.props\n"
+                + "which apply to more dependencies than they should, this can lead to slow builds.\n"
+                + "The following additional pins are recommended:\n"
                 + newLines
                 + "\n\n"
-                + "Run ./gradlew checkOverbroadConstraints --fix to add them.");
+                + "Run ./gradlew checkOverbroadConstraints --fix to add them."
```
CRogers: Needs a newline

src/main/java/com/palantir/gradle/versions/CheckOverbroadConstraints.java
```diff
@@ -97,18 +80,22 @@ private void checkOverbroadConstraints(VersionsProps versionsProps, LockState lo
 
         if (newLines.isEmpty()) {
             return;
-        } else if (getShouldFixProperty().get()) {
+        } else if (getShouldFix().get()) {
```
CRogers: Remove the else here, it's not necessary as the previous thing returned.

src/main/java/com/palantir/gradle/versions/CheckOverbroadConstraints.java
```diff
@@ -0,0 +1,152 @@
...
+            return;
+        } else if (getShouldFix().get()) {
+            getLogger()
+                    .lifecycle("Adding pins to versions.props:\n"
+                            + newLines.stream().collect(Collectors.joining("\n")));
+            writeVersionsProps(getPropsFile().get().getAsFile(), newLines);
+            return;
+        }
+
+        throw new RuntimeException("Over-broad version constraints found in versions.props.\n"
```
CRogers: `ExceptionWithSuggestion` like:

https://github.com/palantir/gradle-consistent-versions/blob/54929cf40b53eb16fd192c1750a91a762740c538/src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsTask.java#L137-L140

CRogers:
Update the changelog to reflect the new name

src/test/java/com/palantir/gradle/versions/CheckOverbroadConstraintsTest.java
```diff
@@ -0,0 +1,60 @@
...
+    private static LockState createLockState(String productionContents) {
+        ConflictSafeLockFile lockReader = new ConflictSafeLockFile(null);
+        return LockState.from(
+                lockReader.parseLines(Stream.of(productionContents.split("\n"))),
+                lockReader.parseLines(Stream.empty()));
+    }
+
+    @Test
+    void test_no_missing_pins_returns_empty() {
+        VersionsProps versionsProps = createVersionProps("com.example.*:*= 1.0.0\ncom.example.core:module = 1.0.1");
```
CRogers: These tests would probably look better with `createVersionsProps` taking a `String... lines` varargs.

src/main/java/com/palantir/gradle/versions/CheckOverbroadConstraints.java
```diff
@@ -0,0 +1,160 @@
...
+
+    public CheckOverbroadConstraints() {
+        getShouldFix().set(false);
+        setGroup(LifecycleBasePlugin.VERIFICATION_GROUP);
+        setDescription(
+                "Ensures matched versions in your versions.lock are pinned to avoid wasted dependency resolution.");
+        getOutputs().upToDateWhen(_task -> true); // task has no outputs, this is needed for it to be up to date
+    }
+
+    @Option(option = "fix", description = "Whether to apply the suggested fix to versions.props")
```
felixdesouza: You can just move this annotation to the `getShouldFix` method.

src/test/java/com/palantir/gradle/versions/CheckOverbroadConstraintsTest.java
FinlayRJW: didnt know you could do this!

src/test/java/com/palantir/gradle/versions/CheckOverbroadConstraintsTest.java
```diff
@@ -0,0 +1,177 @@
...
+        return VersionsProps.fromLines(List.of(lines), null);
+    }
+
+    private static LockState createLockState(String... productionLines) {
+        ConflictSafeLockFile lockReader = new ConflictSafeLockFile(null);
+        return LockState.from(lockReader.parseLines(Stream.of(productionLines)), lockReader.parseLines(Stream.empty()));
+    }
+
+    @Test
+    void test_no_missing_pins_returns_empty() {
```
CRogers: There's no need to prefix every test method with `test`.

FinlayRJW: sorry - old habbits!

src/test/resources/overbroadConstraintsTest/suggest_double_star
```diff
@@ -0,0 +1,9 @@
+versions.props:
+com.*:* = 1.0.0
+
+versions.lock:
+com.example.core:module:1.0.0 (2 constraints: abcdef1)
+com.different.example:artifact:2.0.0 (2 constraints: abcdef1)
+
+new pins (blank when no pins needed):
+com.different.*:* = 2.0.0
```
CRogers: I'm not sure if we ever want to suggest a star in a group, as we really can't be sure if this is right (it's almost certainly not for the software we produce at least). I think perhaps only adding stars in the artifact position makes sense/is enough.

src/test/resources/overbroadConstraintsTest/suggests_star_if_possible
```diff
@@ -0,0 +1,9 @@
+versions.props:
+com.example.*:*= 1.0.0
+
+versions.lock:
+com.example.moduleA:artifact-new:1.0.0 (2 constraints: abcdef1)
+com.example.moduleB:artifact-core:3.0.0 (2 constraints: abcdef1)
+
+new pins (blank when no pins needed):
+com.example.moduleB:* = 3.0.0
```
CRogers: The question here is why did we choose moduleB to add a new pin for rather than moduleA? If I were fixing this manually, I'd probably replace the pin with:

```
com.example.moduleA:* = 1.0.0
com.example.moduleB:* = 3.0.0
```

FinlayRJW: so this comes from moduleA having the same version as the existing pin - currently I am only adding pins and never modifying existing ones. In most examples we could (and maybe should) replace the over-broad pin with a more constrained one

FinlayRJW: I've pushed the changes needed to tighten the lose constraints

src/test/resources/overbroadConstraintsTest/suggests_star_in_complex_situations
```diff
@@ -0,0 +1,17 @@
...
+
+versions.lock:
+com.example.core:artifact-random:1.0.0 (2 constraints: abcdef1)
+com.example.module:artifact-platform-commons:2.0.0 (2 constraints: abcdef1)
+com.example.module:artifact-platform-new:2.0.0 (2 constraints: abcdef1)
+com.example.different:so-very-different:2.0.0 (2 constraints: abcdef1)
+com.example.module:artifact-different:3.0.0 (2 constraints: abcdef1)
+com.example.module:artifact-different-again:4.0.0 (2 constraints: abcdef1)
+
+new pins (blank when no pins needed):
```
CRogers: The new code is better in it's new pin suggestions - but it's deleting half the props in the versions.props when I run the fixer on some internal projects. I think all the ones where the groups in versions.lock items have all the same versions.

I think we should perhaps test a little higher level - have the output here be a diff between the original verisons.props and the new versions.props (I think if you name the file `.diff` github will even syntax highlight it). This way we can test the deleting/inserting code too.

FinlayRJW: which internal projects are causing issues so I can do some quick test

src/main/java/com/palantir/gradle/versions/CheckOverbroadConstraints.java
```diff
@@ -0,0 +1,234 @@
...
+
+    @InputFile
+    public abstract RegularFileProperty getLockFile();
+
+    public CheckOverbroadConstraints() {
+        getShouldFix().set(false);
+        setGroup(LifecycleBasePlugin.VERIFICATION_GROUP);
+        setDescription(
+                "Ensures matched versions in your versions.lock are pinned to avoid wasted dependency resolution.");
+        getOutputs().upToDateWhen(_task -> true); // task has no outputs, this is needed for it to be up to date
```
CRogers: Needs to be `upToDateWhen(_task -> false)` - ie always rerun it. When I was running `--fix` sometimes nothing would happen as I'd already run the task before on the same inputs.

src/test/resources/overbroadConstraintsTest/suggests_star_in_complex_situations
FinlayRJW: Cool - I should have fixed this bug and have ensure that the tests cover it now - I have also switch to the diff method for the tests

src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
```diff
@@ -92,6 +92,15 @@ public final void apply(Project project) {
                     });
             project.getTasks().named("check").configure(task -> task.dependsOn(checkNoUnusedConstraints));
 
+            TaskProvider<CheckOverbroadConstraints> checkBadPins = project.getTasks()
```
CRogers: Change this so we only make the task/have the `check.dependsOn` if there is a gradle property set to true (use `project.findProperty("gcv.overbroad.constraints")` - will need to add to the tests by making a `gradle.properties` file in the setup method and adding `gcv.overbroad.constraints=true`)

src/test/resources/overbroadConstraintsTest/equal_max_types_unwraps_fully.diff
```diff
@@ -0,0 +1,17 @@
...
+com.example:moduleD:2.0.0 (2 constraints: abcdef1)
+com.example:moduleE:3.0.0 (2 constraints: abcdef1)
+
+versions.props diff:
+-com.example:* = 1.0.0
++com.example:moduleE = 3.0.0
++com.example:moduleB = 1.0.0
++com.example:moduleA = 1.0.0
++com.example:moduleD = 2.0.0
++com.example:moduleC = 2.0.0
```
CRogers: Can we try and put them in alphabetical order when inserting?

src/test/groovy/com/palantir/gradle/versions/CheckOverbroadConstraintsIntegrationSpec.groovy
```diff
@@ -0,0 +1,135 @@
...
+        createFile('versions.props')
+        createFile('versions.lock')
+
+        createFile('gradle.properties') << """
+            gcv.overbroad.constraints=true
+        """.stripIndent(true)
+    }
+
+    def buildSucceed() {
+        BuildResult result = runTasks('checkOverbroadConstraints')
```
CRogers, line 36: You can just use `runTasksSuccessfully` here.

src/test/groovy/com/palantir/gradle/versions/CheckOverbroadConstraintsIntegrationSpec.groovy
```diff
@@ -0,0 +1,135 @@
...
+class CheckOverbroadConstraintsIntegrationSpec extends IntegrationSpec {
+
+    def setup() {
+        buildFile << """
+            plugins {
+                id 'com.palantir.versions-lock'
+                id 'com.palantir.versions-props'
+            }
+        """.stripIndent(true)
+        createFile('versions.props')
```
CRogers: We just use `file(...)` normally.

src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
```diff
@@ -92,6 +92,25 @@ public final void apply(Project project) {
                     });
             project.getTasks().named("check").configure(task -> task.dependsOn(checkNoUnusedConstraints));
 
+            String overbroadConstraintsFlag = project.findProperty("gcv.overbroad.constraints") != null
+                    ? project.findProperty("gcv.overbroad.constraints").toString()
+                    : "false";
```
CRogers: I'd probably do `Optional.ofNullable(project.findProperty("gcv.overbroad.constraints")).map(Boolean::parseBoolean).orElse(false)` here to avoid duplication

src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
```diff
@@ -92,6 +92,25 @@ public final void apply(Project project) {
...
+
+            if (Boolean.parseBoolean(overbroadConstraintsFlag)) {
```
CRogers: Actually, new plan. We don't need the feature flag. To test in the background we just need to make the task but not do the `check.dependsOn` bit. This means the task will be there but won't be run unless we explicitly call it.

src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
CRogers: Ignore, see https://github.com/palantir/gradle-consistent-versions/pull/1049/files#r1813043584

src/test/groovy/com/palantir/gradle/versions/CheckOverbroadConstraintsIntegrationSpec.groovy
CRogers: Oh wait, is this the testkit runner?

src/test/java/com/palantir/gradle/versions/CheckOverbroadConstraintsTest.java
```diff
@@ -0,0 +1,322 @@
...
+                    "\n",
+                    "versions.props",
+                    String.join("\n", propsLines),
+                    "",
+                    "versions.lock:",
+                    lockLines,
+                    "",
+                    "versions.props diff:",
+                    diff,
+                    "");
```
CRogers, line 344: I added this to make the diffs nicer. But the current code is always stripping the newline at the end of the versions.props if there is one. We should make sure to preserve the newline if this is there or not. Might be worth writing this as a check that runs on every single test rather than depending on PR review of the examples.

src/test/java/com/palantir/gradle/versions/CheckOverbroadConstraintsTest.java
```diff
@@ -0,0 +1,322 @@
...
+        }
+
+        public static Builder builder(String testName) {
+            return new Builder(testName);
+        }
+
+        public List<String> newPropsLines() {
+            Map<String, List<String>> oldToNewLines =
+                    CheckOverbroadConstraints.determineNewLines(versionsProps, lockState);
+            return CheckOverbroadConstraints.generateUpdatedPropsLines(oldToNewLines, propsLines);
```
CRogers: We should run the check overbroad constraints on the new versions.props just to verify it does actually fix the problem for each of the tests (don't want to have a case where the fix does not fully fix the issue and the check fails after applying it).

src/test/groovy/com/palantir/gradle/versions/CheckOverbroadConstraintsIntegrationSpec.groovy
FinlayRJW: I'm not sure that the testkit runner is but this seem to be how all the groovy tests are done in gcv

src/test/java/com/palantir/gradle/versions/CheckOverbroadConstraintsTest.java
FinlayRJW: So I think what is happening here is that diff utils is removing the new line as it is shared between the old and new versions props - what I have done is add some specific test case that only check generateUpdatedPropsLines they should cover checking new lines

src/test/java/com/palantir/gradle/versions/CheckOverbroadConstraintsTest.java
FinlayRJW: This was a really good idea and uncovered a bug where as we didn't remove matched artifacts from the unmatched list we were ending up with multiple matches of the same artifact if there was a broader constraint - I have fixed this and added a check to ensure on a second run there are no new lines

src/test/java/com/palantir/gradle/versions/CheckOverbroadConstraintsTest.java
CRogers: It's removing the newline unnecessarily on internal projects too. Trust me, angery people will appear in our support when we excavate out newline changes... (I have done this many times).

FinlayRJW: ah! The culprit was the read from file function not adding the new line to the list - so the test I added wouldn't catch it - Ive modified the groovy tests to include checking for new line

CRogers:
:+1:

PR #1053, opened 2023-05-29
Improvement: Upgradle to 7.6.1

robert3005:
## Before this PR
Currently this repo is on 7.5.1 and will not automatically upgarde to 7.6.1

## After this PR

==COMMIT_MSG==
Upgrade gradle to 7.6.1
==COMMIT_MSG==
This is done as a step towards gradle 8 compatibility

PR #1060, opened 2023-06-23
Improvement: Add gradle 8 compatibility

robert3005:
## Before this PR
We couldn't run against gradle 8

## After this PR

==COMMIT_MSG==
Support gradle 8
==COMMIT_MSG==

There's deprecations to be resolved before gradle 9. I think they boil down to completely purging runtime and compile classpath configurations from referneces and using apiElements and runtimeElements, however, in my limited attempts naive swap didn't work so it might need deeper investigation

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.6-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.6.1-bin.zip
 networkTimeout=10000
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
--- a/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
@@ -661,8 +661,6 @@ public class VersionsLockPlugin implements Plugin<Project> {
             copiedConf.getAttributes().attribute(Usage.USAGE_ATTRIBUTE, internalUsage);
             // Must set this because we depend on this configuration when resolving unifiedClasspath.
             copiedConf.setCanBeConsumed(true);
-            // But this should never be resolved! (it will most likely fail to given the usage above)
-            copiedConf.setCanBeResolved(false);
             // Since we only depend on these from the same project (via CONSISTENT_VERSIONS_PRODUCTION or
             // CONSISTENT_VERSIONS_TEST), we shouldn't allow them to be visible outside this project.
             copiedConf.setVisible(false);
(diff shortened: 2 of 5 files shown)


src/test/groovy/com/palantir/gradle/versions/ConsistentVersionsPluginIntegrationSpec.groovy
```diff
@@ -115,34 +115,6 @@ class ConsistentVersionsPluginIntegrationSpec extends IntegrationSpec {
         gradleVersionNumber << GRADLE_VERSIONS
     }
 
-    // TODO(dsanduleac): should remove this since this functionality doesn't fully work anyway, and we are
```
robert3005, line 118: This test was redundant since it stopped testing `compile` configuration as a result of automated migration.

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -661,8 +661,6 @@ private void recursivelyCopyProjectDependenciesWithScope(
             copiedConf.getAttributes().attribute(Usage.USAGE_ATTRIBUTE, internalUsage);
             // Must set this because we depend on this configuration when resolving unifiedClasspath.
             copiedConf.setCanBeConsumed(true);
-            // But this should never be resolved! (it will most likely fail to given the usage above)
```
robert3005, line 664: This is the real change to make it work. You can't alter `resolvable` after configuration has been created. If configuration is resolvable the clone will be and there's no way of changing it.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.6-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.6.1-bin.zip
```
robert3005, line 3: this got downgraded by an automated pr

robert3005:
@fawind this is a follow up to #1053 so that we support gradle 8

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
CRogers: Thanks for the fix! This sounds ok to me (and I guess we have no choice) but GCV changes are always exciting in unexpected ways. I'm going to release an RC and check this works out fine on a few internal repos.

CRogers: https://github.com/palantir/gradle-consistent-versions/releases/tag/2.14.0-rc1

gradle/wrapper/gradle-wrapper.properties
CRogers: I've bumped the version in excavator so this shouldn't happen again.

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
CRogers: Had to make a branch here so we can make an RC release: https://github.com/palantir/gradle-consistent-versions/pull/1064. Gonna close this PR in favour of that one.

robert3005:
thanks a lot. I do agree it can be surprising. It took me a bit of time to figure out why that 1 line change was necessary

CRogers:
Closing in favour of #1064.

gradle/wrapper/gradle-wrapper.properties
CRogers: Actually, I've scratched that due to https://github.com/gradle/gradle/issues/24234.

robert3005: ah, good to know

PR #1077, opened 2023-08-24
Fix publish to gradle plugin portal

CRogers:
## Before this PR
GCV 2.13.0 is not available in the gradle plugin portal - [the latest version is 2.12.0](https://plugins.gradle.org/plugin/com.palantir.consistent-versions).

There was [a publish failure for 2.13.0](https://app.circleci.com/pipelines/github/palantir/gradle-consistent-versions/2968/workflows/491ef2c9-cd57-404c-b0df-fc8909590f0c/jobs/13231):

```
Caused by: java.lang.IllegalArgumentException: Plugin 'versionsLock' has no 'displayName' property set
	at com.gradle.publish.AbstractConfigValidator.validateProperty(AbstractConfigValidator.java:137)
	at com.gradle.publish.AbstractConfigValidator.validateDisplayName(AbstractConfigValidator.java:124)
```

## After this PR
==COMMIT_MSG==
Fix publishing to gradle plugins portal.
==COMMIT_MSG==

... by adding `displayName` property. And ensure we set `description` too in case it becomes required.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -91,19 +91,21 @@ gradlePlugin {
         consistentVersions {
             id = 'com.palantir.consistent-versions'
             implementationClass = 'com.palantir.gradle.versions.ConsistentVersionsPlugin'
-            description = 'Compact, constraint-friendly lockfiles for your dependencies'
             displayName = 'Compact, constraint-friendly lockfiles for your dependencies'
+            description = displayName
             tags.set(['versions'])
         }
         versionsLock {
             id = 'com.palantir.versions-lock'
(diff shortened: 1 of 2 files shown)


felixdesouza (approved):

CRogers:
whut

felixdesouza:
oh no, did we do the rc off of develop?

CRogers:
> Caused by: java.lang.IllegalArgumentException: Plugin 'versionsLock' has no 'tags' property set

Why oh why does this verification not happen in `check`, only in the task that actually publishes it???

Issue #691, opened 2021-03-26
Scope not recognized for dependencies in testFixtures*


dweiss:
Dependencies in testFixturesApi (and other configurations), now standard, are not recognized and cause this:
```
> Couldn't determine scope for dependency: bzzzz...:1.0.4
```

swanysimon:
Related: https://github.com/palantir/gradle-consistent-versions/issues/308

I tried to force include the `testFixtures` source set, but this can cause issues with mismatched versions:

```groovy
versionsLock {
    test {
        from(sourceSets.testFixtures)
    }
}
```

My current strategy is to make `integrationTest` extend from `test`, and then declare a `testImplementation` dependency on every dependency in a `testFixtures*` configuration. It's extremely unfortunate.

carterkozak:
Hmm, the plugin will include all source sets with a name ending `test` (case insensitive) e.g. `test` and `integrationTest`, but I suppose it would be reasonable to include those which start with `test` as well. If you can rename `testFixtures` to end with `test`, that will unblock using existing versions.

https://github.com/palantir/gradle-consistent-versions/blob/a2212cfc89e01f16458aaee4df6a2440ec644520/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java#L992-L996

swanysimon:
The `testFixtures*` configurations are from a [core Gradle plugin](https://docs.gradle.org/current/userguide/java_testing.html#sec:java_test_fixtures), so not really an option for us. I do like the idea of looking for configurations starting with `test`, since I (currently) can't think of any examples where you wouldn't want those included in the test version locks.

Issue #308, opened 2019-10-18
Comaptibility with java-test-fixtures


juriad:
## What happened?

Using **gradle 5.6.2** applying both this plugin and **java-test-fixtures** ([new in 5.6](https://docs.gradle.org/current/userguide/java_testing.html#sec:java_test_fixtures)), I receive an error when **writing locks** due to unresolved dependencies.

```
Caused by: org.gradle.api.GradleException: Could not compute lock state from configuration 'unifiedClasspath' due to unresolved dependencies:
 * project :p1 (requested: 'project :p1' because: requested)
      Failures:
         - Could not resolve project :p1.
         - Module 'com.mycompany:p1' has been rejected:
[...]

PR #1118, opened 2024-01-22
Adds support for use with java-test-fixtures, a core Gradle plugin.

Kennelly57:
## Before this PR

Currently, there is no support for using  java-test-fixtures, a core Gradle plugin. When using a test fixture, this plugin will fail. As an example, here is an error message that will be logged:

`Couldn't determine scope for dependency: com.h2database:h2:2.2.224`

if the build.gradle file is as follows:
```groovy
dependencies {
    testFixturesApi 'com.h2database:h2'
}
```

### Related Issues
https://github.com/palantir/gradle-consistent-versions/issues/691
https://github.com/palantir/gradle-consistent-versions/issues/308

## After this PR

==COMMIT_MSG==
Adds support for use with java-test-fixtures, a core Gradle plugin.
==COMMIT_MSG==

A case to check for test sourceSets starting with the word "test" has been added, as suggested in #691 by @carterkozak. I have also included a test case to check that the lock file is properly generated when using the testFixtures plugin.

## Possible downsides?

The logic is dead-simple, so it is unlikely to cause unforeseen bugs. However, this now means that the plugin may change some sourceSets from production to test. I imagine that this impact will be small because it is unlikely many people are labelling their production source sets with test as a prefix. If the impact is deemed too high, then I also would be happy just to scan for the prefix 'testFixture'

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,8 +137,8 @@ com.jayway.awaitility:awaitility:1.6.5 (1 constraints: c615c1d2)
 ```
 
 The lockfile sources production dependencies from the _compileClasspath_ and _runtimeClasspath_ configurations, and
-test dependencies from the compile/runtime classpaths of any source set named `jmh` or whose name ends in "test"
-(e.g. `test`, `integrationTest`, `eteTest`).
+test dependencies from the compile/runtime classpaths of any source set named `jmh` or whose name starts with or ends 
+in "test" (e.g. `test`, `integrationTest`, `eteTest`).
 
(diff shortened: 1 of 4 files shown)


palantirtech:
Thanks for your interest in [palantir/gradle-consistent-versions](https://github.com/palantir/gradle-consistent-versions), @Kennelly57! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #1129, opened 2024-02-13
Document versions.props comment format

ash211:
## Before this PR
Expected `versions.props` comment format is not documented: https://github.com/palantir/gradle-consistent-versions?tab=readme-ov-file#versionsprops-lower-bounds-for-dependencies

## After this PR
==COMMIT_MSG==
Document current `versions.props` comment format
==COMMIT_MSG==

This also adds test coverage for the documented comment formats, plus some other comment behaviors that currently exist.  I'm not sure the test case in `ignores_comment_on_same_line_with_no_hash()` is ideal, but supporting that behavior doesn't seem to be hurting much.

## Possible downsides?
None known

### DIFF
--- a/src/test/groovy/com/palantir/gradle/versions/VersionsPropsTest.java
+++ b/src/test/groovy/com/palantir/gradle/versions/VersionsPropsTest.java
@@ -93,9 +93,23 @@ public class VersionsPropsTest {
     }
 
     @Test
+    void ignores_comment_on_same_line_with_hash_and_no_spaces() throws IOException {
+        Path propsFile = tempDir.resolve("versions.props");
+        Files.writeString(
+                propsFile,
+                "com.palantir.test:test = 1.0.0#comment on same line with hash and no spaces",
+                StandardCharsets.UTF_8);
+
+        VersionsProps versionsProps = VersionsProps.loadFromFile(propsFile);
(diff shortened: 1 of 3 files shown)


src/test/groovy/com/palantir/gradle/versions/VersionsPropsTest.java
```diff
@@ -59,4 +59,60 @@ void fails_to_load_illegal_version() throws IOException {
...
+        VersionsProps versionsProps = VersionsProps.loadFromFile(propsFile);
+        assertThat(versionsProps.getFuzzyResolver().exactMatches()).containsExactly("com.palantir.test:test");
+        assertThat(versionsProps.getFuzzyResolver().globs()).isEmpty();
+    }
+
+    @Test
+    void ignores_comment_on_same_line_with_no_hash() throws IOException {
+        Path propsFile = tempDir.resolve("versions.props");
+        Files.writeString(
+                propsFile, "com.palantir.test:test = 1.0.0  comment on same line with no hash", StandardCharsets.UTF_8);
```
ash211, line 88: this seems less than ideal, but I'm not sure we gain much by making this invalid

ash211:
Still valuable but only P3

PR #1233, opened 2024-09-17
Plugin to create a maven repo list in the .ideas folder

FinlayRJW:
## Before this PR
Had no way of generating a list of repos for the intellij version props plugin to use

## After this PR
We can now use this plugin to generate the repo list

## Possible downsides?
This will create the list even if you are not using the intellij plugin

**Had to bump minimum gradle version to `7.6.4`** this means `7.2` is no longer supported


src/main/java/com/palantir/gradle/versions/VersionsPropsIdeaPlugin.java
```diff
@@ -0,0 +1,96 @@
...
+                .map(url -> url.endsWith("/") ? url : url + "/")
+                .distinct()
+                .collect(Collectors.toList());
+
+        File ideaDir = new File(target.getRootProject().getProjectDir(), ".idea");
+
+        writeRepositoryUrlsToXml(repositoryUrls, new File(ideaDir, "mavenRepositories.xml"));
+    }
+
+    private static void writeRepositoryUrlsToXml(List<String> repositoryUrls, File file) {
```
felixdesouza: It's not immediately clear what the structure of said repository file is without having to run this code in my head. We typically use Jackson (has xml support) together with [immutables](https://immutables.github.io/) (has nice builders). Depending on whether or not you want to couple the apis, you can even use the same object to deserialise it within IntelliJ!

You can do something like:

```gradle
dependencies {
  // assuming you have usual jackson dependencies already
  implementation 'com.fasterxml.jackson.dataformat:jackson-dataformat-xml'
}
```

```java
@Value.Immutable
@JsonDeserialize(as = ImmutableMavenRepositories.class)
@JsonSerialize(as = ImmutableMavenRepositories.class)
@JsonRootName("repositories")
interface MavenRepositories {
  @Value.Parameter
  @JacksonXmlElementWrapper(useWrapping = false)
  @JacksonXmlProperty(localName = "repository")
  List<MavenRepository> repository();
}

@Value.Immutable
@JsonDeserialize(as = ImmutableMavenRepository.class)
@JsonSerialize(as = ImmutableMavenRepository.class)
interface MavenRepository {
  @Value.Parameter
  @JacksonXmlProperty(isAttribute = true)
  String value();
}
```

So to actually do what you want to do:
```java
// make all the objects
Set<String> repositoryUrls = ...; 
// we can actually just combine it here, but I don't want to copy for brevity sake

repositoryUrls.stream()
  .map(ImmutableMavenRepository::of)
  .collect(Collectors.collectingAndThen(Collectors.toList(), ImmutableMavenRepositories::of);
```

In case you're wondering what [`@Value.Parameter`](https://immutables.github.io/immutable.html#constructor-method) does.

It's not strictly necessary, it'd just be a bit more cumbersome:

```java
public static MavenRepository of(String url) {
  return ImmutableMavenRepository.builder()
    .value(url)
    .build();
}
```

## How to write this out?
The above annotations:
* `@JsonDeserialise`
* `@JsonSerialise`
* `@JsonRootName`
* `@JacksonXmlProperty`

are annotations that are part of jackson. Jackson is the serialisation/deserialisation library for many different formats i.e. json, yaml, and in this case xml. They're controlling what jackson is supposed to do on both serialisation and deserialisation.

Once you have an object, you need to serialise it:

```java
XmlMapper mapper = new XmlMapper();

// typically we use ObjectMapper for both yaml and json, 
// but XmlMapper is needed here for xml related things
MavenRepositories mavenRepositories = ...;
mapper.writerWithDefaultPrettyPrinter().writeValue(file, mavenRepositories);
```

If you have an xml file and you want it read it into a nice to use object without having to go through manual document and element apis, you can use the opposite apis:

```java
XmlMapper mapper = new XmlMapper();
MavenRepositories mavenRepositories = mapper.readValue(file, MavenRepositories.class);
```

Yes, the above is more code, but it's easy to refer to what a document might look like without having to run it. It also means you can do things via higher level operations i.e. make an object, and write it out as opposed to make the  right invocations in the right order for the low-level XML writer.

src/test/groovy/com/palantir/gradle/versions/VersionPropsIdeaPluginTest.groovy
```diff
@@ -0,0 +1,54 @@
...
+
+        System.setProperty('idea.active', 'true')
+    }
+
+    def 'plugin creates mavenRepositories.xml file in .idea folder'() {
+        when:
+        runTasksSuccessfully('idea')
+
+        then:
+        def repoFile = new File(projectDir, '.idea/mavenRepositories.xml')
```
felixdesouza: with the above change, your tests become a lot nicer, because you can do something like:

```groovy
def repositories = xmlMapper.readValue(repoFile, MavenRepositories.class);
def expectedRepositories = ImmutableMavenRepositories.builder()
  .addRepository(ImmutableMavenRepository.of("https://test"))
  .addRepository(ImmutableMavenRepository.of("https://demo/"))
  .build();

repositories == expectedRepositories
```

src/test/groovy/com/palantir/gradle/versions/VersionPropsIdeaPluginTest.groovy
```diff
@@ -0,0 +1,54 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.gradle.versions
+
+import nebula.test.IntegrationSpec
+
+class VersionPropsIdeaPluginTest extends IntegrationSpec {
+    def setup() {
```
felixdesouza: use [`@RestoreSystemProperties`](https://spockframework.org/spock/docs/2.3/extensions.html#_restoresystemproperties) for test hygiene

felixdesouza:
An aside, and not a blocker, but I'm curious about making use of the [Tooling API](https://docs.gradle.org/current/userguide/third_party_integration.html) here / with intellij-gradle integration generally. From what I've understood, a gradle plugin (or init script, how intellij does it), can register a `ToolingModelBuilder` that will build a model that can then be queried later on within something like IntelliJ. This is how IntelliJ does it to get the [projects](https://github.com/JetBrains/intellij-community/blob/master/plugins/gradle/tooling-extension-impl/src/com/intellij/gradle/toolingExtension/impl/model/projectModel/GradleExternalProjectModelBuilder.java), dependencies and a bunch of other pieces. 

Thoughts @CRogers

src/test/groovy/com/palantir/gradle/versions/VersionPropsIdeaPluginTest.groovy
FinlayRJW: I couldn't seem to apply this to the setup method but it did seem to work on the actual test - is that correct

src/main/java/com/palantir/gradle/versions/VersionsPropsIdeaPlugin.java
FinlayRJW: thanks for the review this is super detailed! I have now switched to jackson :)

FinlayRJW:
for some reason adding `implementation 'com.fasterxml.jackson.dataformat:jackson-dataformat-xml'` has broken a lot of tests due to 
```
.gradle-test-kit/caches/jars-9/507792a4bf4466b62991b4bdf628dc95/jackson-core-2.17.2.jar.
	at org.gradle.internal.classpath.ClasspathBuilder.jar(ClasspathBuilder.java:55)
	at org.gradle.internal.classpath.InstrumentingClasspathFileTransformer.instrument(InstrumentingClasspathFileTransformer.java:137)
	at org.gradle.internal.classpath.InstrumentingClasspathFileTransformer.transform(InstrumentingClasspathFileTransformer.java:132)
	at org.gradle.internal.classpath.InstrumentingClasspathFileTransformer.transform(InstrumentingClasspathFileTransformer.java:96)
	at org.gradle.internal.classpath.DefaultCachedClasspathTransformer.transformFile(DefaultCachedClasspathTransformer.java:193)
	at org.gradle.internal.classpath.DefaultCachedClasspathTransformer.lambda$cachedFile$6(DefaultCachedClasspathTransformer.java:186)
	at org.gradle.internal.UncheckedException.unchecked(UncheckedException.java:74)
	at org.gradle.internal.classpath.DefaultCachedClasspathTransformer.lambda$transformAll$9(DefaultCachedClasspathTransformer.java:236)
	at org.gradle.internal.concurrent.ExecutorPolicy$CatchAndRecordFailures.onExecute(ExecutorPolicy.java:64)
	at org.gradle.internal.concurrent.ManagedExecutorImpl$1.run(ManagedExecutorImpl.java:48)
	at org.gradle.internal.concurrent.ThreadFactoryImpl$ManagedThreadRunnable.run(ThreadFactoryImpl.java:61)
Caused by: java.lang.IllegalArgumentException: Unsupported class file major version 65
```

build.gradle
```diff
@@ -113,6 +115,13 @@ gradlePlugin {
             description = displayName
             tags.set(['versions'])
         }
+        VersionsPropsIdeaPlugin {
```
CRogers: Should probably be lower case v

src/main/java/com/palantir/gradle/versions/MavenRepositories.java
```diff
@@ -0,0 +1,36 @@
...
+import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlElementWrapper;
+import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlProperty;
+import java.util.List;
+import org.immutables.value.Value;
+
+@Value.Immutable
+@JsonDeserialize(as = ImmutableMavenRepositories.class)
+@JsonSerialize(as = ImmutableMavenRepositories.class)
+@JsonRootName("repositories")
+interface MavenRepositories {
```
CRogers: I expect we're going to want to use this code to parse the XML in the gradle-consistent-versions-idea-plugin repo... unless we end up parsing it using some intellij mechanism. If so, it would make sense to put this in it's own project so the idea-plugin can include just this classes/dependencies

src/main/java/com/palantir/gradle/versions/VersionsPropsIdeaPlugin.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.stream.Collectors;
+import org.gradle.api.Plugin;
+import org.gradle.api.Project;
+import org.gradle.api.artifacts.repositories.MavenArtifactRepository;
+
+public final class VersionsPropsIdeaPlugin implements Plugin<Project> {
+
+    @Override
+    public void apply(Project target) {
+        target.getPluginManager().withPlugin("idea", _ideaPlugin -> {
```
CRogers: I'm not sure the `idea` plugin gets applied when using the built in intellij integration. Many projects apply it still, but it's not necessarily required (I think).

src/main/java/com/palantir/gradle/versions/VersionsPropsIdeaPlugin.java
```diff
@@ -0,0 +1,61 @@
...
+            try {
+                configureIntellij(target);
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
+        });
+    }
+
+    private static void configureIntellij(Project target) throws IOException {
+        if (!Boolean.getBoolean("idea.active")) {
```
CRogers, line 35: Might as well move this to the `apply` method.

src/main/java/com/palantir/gradle/versions/VersionsPropsIdeaPlugin.java
```diff
@@ -0,0 +1,61 @@
...
+            return;
+        }
+        Set<String> repositoryUrls = target.getRootProject().getAllprojects().stream()
+                .flatMap(project -> project.getRepositories().withType(MavenArtifactRepository.class).stream())
+                .map(MavenArtifactRepository::getUrl)
+                .map(Object::toString)
+                .map(url -> url.endsWith("/") ? url : url + "/")
+                .collect(Collectors.toSet());
+
+        File file = new File(target.getRootProject().getProjectDir(), ".idea/mavenRepositories.xml");
```
CRogers: I'd probably call this something like `gcv-maven-repositories.xml` to make it clear it's related to gcv.

src/main/java/com/palantir/gradle/versions/MavenRepositories.java
CRogers: It might be worth looking at how palantir-java-format does this actually (look at `.idea/palantir-java-format.xml`). It makes a plugin settings page for itself and we configure that file. It would expose the repos in the UI, which idk is good or bad.

src/main/java/com/palantir/gradle/versions/VersionsPropsIdeaPlugin.java
FinlayRJW: renamed to `gradle-consistent-versions-plugin-settings.xml` as am now creating a settings xml

src/main/java/com/palantir/gradle/versions/VersionsPropsIdeaPlugin.java
FinlayRJW: do you mean that we can call `configureIntellij(project);`  without wrapping it in `target.getPluginManager().withPlugin("idea"`?

src/main/java/com/palantir/gradle/versions/MavenRepositories.java
FinlayRJW: So after looking into this I have switched to creating / updating a idea plugin settings xml that can now be easily accessed and modified from within plugin - this has let me add a project level disable button for projects that require it see https://github.com/palantir/gradle-consistent-versions-idea-plugin/pull/27

FinlayRJW:
I've had to make this run `afterEvaluate` to ensure that the repos are correctly collected is there an earlier point this could be run?

FinlayRJW:
I have now moved this to be run in a task so it is no longer anything todo with the plugin settings

src/main/java/com/palantir/gradle/versions/VersionsPropsIdeaPlugin.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    public void apply(Project project) {
+
+        if (!Boolean.getBoolean("idea.active")) {
+            return;
+        }
+
+        TaskProvider<GenerateMavenRepositoriesTask> writeMavenRepositories = project.getTasks()
+                .register("writeMavenRepositories", GenerateMavenRepositoriesTask.class, task -> {
+                    Provider<Set<String>> repositoryProvider = project.provider(() -> project.getRepositories().stream()
```
CRogers: I would just inline this provider rather than make a variable for it.

src/test/groovy/com/palantir/gradle/versions/VersionPropsIdeaPluginIntegrationSpec.groovy
```diff
@@ -0,0 +1,84 @@
...
+            }
+
+            apply plugin: 'com.palantir.version-props-idea'
+            apply plugin: 'idea'
+        """.stripIndent(true)
+
+        def ideaDir = new File(projectDir, '.idea')
+        ideaDir.mkdirs()
+
+        System.setProperty('idea.active', 'true')
```
CRogers: I think it's best not to test like this using `System.setProperty` as it will set the property on all further tests. `@RestoreSystemProperties` is a bit dangerous as you have to remember to add it to each test. Instead, you can set the system property as part of the gradle invocation:

```
runTasksSuccessfully('-Didea.active=true')
```

CRogers:
Looks good given the comments.

CRogers (approved):

PR #1234, opened 2024-09-17
Intellij Plugin

FinlayRJW:
## Before this PR
Previously we had no way of handling version.props files in intellij

## After this PR
Adds a intellij plugin with the following features for version.props files:
- code highlighting
- code commenting
- auto suggest group package and version
- auto run wVL on save

## Possible downsides?
currently hardcoded for just maven central - can be laggy with slow repos


idea-plugin/build.gradle
```diff
@@ -0,0 +1,44 @@
...
+intellij {
+    pluginName = "gradle-consistent-versions"
+    updateSinceUntilBuild = true
+    version = "2024.1"
+    plugins = ['java', 'org.jetbrains.plugins.gradle']
+}
+
+patchPluginXml {
+    pluginDescription = "Language plugin for version.props allowing for auto-lock updating"
+    version = "1.0.0" // For some reason project.version is unspeificed and thus this fails
```
FinlayRJW: Not sure why am needing to hard code this currently @CRogers any ideas?

CRogers: https://github.com/palantir/gradle-consistent-versions/pull/1235

idea-plugin/build.gradle
```diff
@@ -0,0 +1,44 @@
...
+    sinceBuild = '241' // TODO: test against this version of IntelliJ to ensure no regressions
+    untilBuild = ''
+}
+
+afterEvaluate {
+    tasks.named('test').configure { it.javaLauncher = null }
+}
```
CRogers: Should move this to the external-publish-plugin.

idea-plugin/build.gradle
```diff
@@ -0,0 +1,44 @@
...
+
+// Javadoc fails if there are no public classes to javadoc, so make it stop complaining.
+tasks.named("javadoc", Javadoc) {
+    failOnError = false
+}
+
+tasks.withType(JavaCompile).configureEach {
+    options.errorprone.disable 'PreferSafeLoggableExceptions'
+    options.errorprone.disable 'PreferSafeLogger'
```
CRogers: Might want to do this [in baseline like we do here for java-gradle-plugin](https://github.com/palantir/gradle-baseline/blob/fd590105ee4f4d8e746224d680f3bf8d557ab0d0/gradle-baseline-java/src/main/groovy/com/palantir/baseline/plugins/BaselineErrorProne.java#L100-L114) (except for intellij)

idea-plugin/src/main/generated/com/palantir/gradle/versions/intellij/VersionPropsLexer.java
CRogers, line 1: We should try not to check in the generated code. It should really be generated at build time...

idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/DependencyGroup.java
```diff
@@ -0,0 +1,74 @@
...
+        if (url.contains("IntellijIdeaRulezzz")) {
+            return "";
+        }
+        if (!url.isEmpty()) {
+            url = url + "/";
+        }
+        return url;
+    }
+
+    public static DependencyGroup groupFromParameters(@NotNull CompletionParameters parameters) {
```
CRogers: We don't ever use the `@NotNull` annotations, we prefer to use uber's https://github.com/uber/NullAway and take everything that isn't `@Nullable` to be not null.

idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
```diff
@@ -0,0 +1,153 @@
...
+import org.immutables.value.Value;
+import org.jsoup.Jsoup;
+import org.jsoup.nodes.Document;
+import org.jsoup.nodes.Element;
+import org.jsoup.select.Elements;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.w3c.dom.NodeList;
+
+@Value.Immutable
```
CRogers: This doesn't feel like the kind of type that should be an immutable - it's never going to be compared against another object using hashCode or equals. I'd make it a "normal" type and just not implement hashcode and equals. Generally types ending in `-er` are never not going to be value types.

idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/VersionPropsCodeInsightTest.java
```diff
@@ -0,0 +1,61 @@
...
+ * limitations under the License.
+ */
+package com.palantir.gradle.versions.intellij;
+
+import com.intellij.codeInsight.completion.CompletionType;
+import com.intellij.testFramework.fixtures.LightJavaCodeInsightFixtureTestCase;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+public class VersionPropsCodeInsightTest extends LightJavaCodeInsightFixtureTestCase {
```
CRogers, line 23: Nice! Actual tests.

idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionCompletionContributor.java
```diff
@@ -0,0 +1,77 @@
...
+                        VersionPropsDependencyVersion versionElement =
+                                ReadAction.compute(() -> (VersionPropsDependencyVersion)
+                                        parameters.getPosition().getParent());
+
+                        VersionPropsProperty property = ReadAction.compute(() -> findParentProperty(versionElement));
+
+                        DependencyGroup group = DependencyGroup.fromString(
+                                property.getDependencyGroup().getText());
+                        DependencyPackage dependencyPackage = DependencyPackage.of(
+                                property.getDependencyName().getText());
```
CRogers: We should be consistent with the naming - I'd call `DependencyPackage` `DependencyName` instead. To me package means group and name together.

idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/VersionPropsCodeInsightTest.java
```diff
@@ -0,0 +1,61 @@
...
+    @Override
+    protected final String getTestDataPath() {
+        return "src/test/testData";
+    }
+
+    @Test
+    public void test_version_completion() throws Exception {
+        setUp();
+        myFixture.configureByFiles("VersionTestData.props");
```
felixdesouza: are you able to use `configureByText` instead? If it's not a large amount of text, it makes it easier to process the test as you don't have to jump around to find the underlying file. Which in this case is not as straightforward as you actually have to plumb through IntelliJ's test helpers

idea-plugin/src/main/generated/com/palantir/gradle/versions/intellij/VersionPropsLexer.java
FinlayRJW: I have switched to using the gradle grammar kit plugin to generate the code at build :)

idea-plugin/build.gradle
FinlayRJW: awesome - merged that in and it seems to be working!

idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/DependencyGroup.java
FinlayRJW: good to know thanks!

idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/RepositoryExplorer.java
FinlayRJW: thanks - have turned back to a regular class

idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/VersionPropsCodeInsightTest.java
FinlayRJW: Quite pleased with these, quite basic at the moment but can expand coverage in the future. Currently I am using real packages in maven central for the tests - is this a bad idea? would I be better of mocking this so I'm not relying on maven and the packages not changing

idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionCompletionContributor.java
FinlayRJW: good point, not sure why I switched half way through! have refactored

idea-plugin/src/test/java/com/palantir/gradle/versions/intellij/VersionPropsCodeInsightTest.java
FinlayRJW: switched to configure by text thanks!

idea-plugin/build.gradle
FinlayRJW: done see https://github.com/palantir/gradle-external-publish-plugin/pull/460

idea-plugin/build.gradle
FinlayRJW: added to baseline - see https://github.com/palantir/gradle-baseline/pull/2878

idea-plugin/build.gradle
```diff
@@ -0,0 +1,56 @@
...
+}
+
+generateLexer {
+    sourceFile = file('src/main/java/com/palantir/gradle/versions/intellij/VersionProps.flex')
+    targetOutputDir = file('src/main/generated/com/palantir/gradle/versions/intellij')
+    targetClass = 'com.palantir.gradle.versions.intellij.VersionPropsLexer'
+    purgeOldFiles = true
+}
+
+compileJava.dependsOn generateParser, generateLexer
```
CRogers: Mmm, this is better now we can generate the code in gradle.

Just a small touchup to this line. It's best to at least prefix the task names with `tasks.`. Otherwise this can match extensions, functions, other things that Gradle magically places in global scope:

```gradle
tasks.compileJava.dependsOn tasks.generateParser, tasks.generateLexer
```

This actually forces the creation of these tasks at configuration time, which can be a performance issue. The absolute "best" way of writing this is:

```gradle
tasks.named('compileJava').configure {
    dependsOn tasks.named('generatedParser'), tasks.named('generateLexer')
}
```

which is as lazy as possible, but is maybe a bit verbose for such a small repo like this.

idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/DependencyGroup.java
```diff
@@ -0,0 +1,68 @@
...
+
+    public static DependencyGroup fromString(String string) {
+        return ImmutableDependencyGroup.of(Arrays.asList(string.split("\\.")));
+    }
+
+    public final String asUrlString() {
+        String url = String.join("/", parts());
+
+        // For some odd reason if you edit the first part of the group you get "IntellijIdeaRulezzz"
+        if (url.contains("IntellijIdeaRulezzz")) {
```
CRogers: It feels weird to have this check at this level. If a `CompletionContributor` can always add this, we probably want to add some code there to stop it rather than than here?

CRogers: I think this just appears if the set of completions is empty for some reason:

> [A default string that is inserted to the file before completion to guarantee that there'll always be some non-empty element there](https://github.com/JetBrains/intellij-community/blob/528b777e4a780350d7c8b0ccd1574e9102ccb71b/platform/core-api/src/com/intellij/codeInsight/completion/CompletionUtilCore.java#L9)

CRogers: Perhaps `groupFromParameters` should return an `Optional<DependencyGroup>`, rather than returning an empty string here that the caller has to remember to check.

idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GroupCompletionContributor.java
```diff
@@ -0,0 +1,57 @@
...
+
+        extend(
+                CompletionType.BASIC,
+                PlatformPatterns.psiElement(VersionPropsTypes.GROUP_PART),
+                new CompletionProvider<>() {
+                    @Override
+                    protected void addCompletions(
+                            @NotNull CompletionParameters parameters,
+                            @NotNull ProcessingContext context,
+                            @NotNull CompletionResultSet resultSet) {
```
CRogers: More NotNulls

idea-plugin/build.gradle
FinlayRJW: Have updated this now

idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/DependencyGroup.java
FinlayRJW: This was caused by me forgetting to check if the current element was the name or group, this meant when you edited the first group part in a completed line it would incorrectly trigger the name logic, Ive now fixed this so this check isnt needed anymore

idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/GroupCompletionContributor.java
FinlayRJW: thinks Ive got them all now

.gitignore
```diff
@@ -39,5 +39,8 @@ reports/
 *.tmproject
 *.sw[op]
 
+# Intellij generated code
+idea-plugin/src/main/generated
```
CRogers, line 43: Do we control where this is outputted to?

idea-plugin/build.gradle
```diff
@@ -0,0 +1,59 @@
...
+}
+
+grammarKit {
+    jflexRelease = "1.9.2"
+    grammarKitRelease = "2022.3.2"
+}
+
+generateParser {
+    sourceFile = file('src/main/java/com/palantir/gradle/versions/intellij/VersionProps.bnf')
+    targetRootOutputDir = file('src/main/generated')
```
CRogers: I think we should put the generate files in the build directory. The convention is in `build/generated/sources/intellij-parser/main/java`

CRogers: This means you can remove the line from the .gitignore too.

idea-plugin/build.gradle
```diff
@@ -0,0 +1,59 @@
...
+    pathToParser = 'com/palantir/gradle/versions/intellij/parser/VersionPropsParser'
+    pathToPsiRoot = 'com/palantir/gradle/versions/intellij/psi'
+    purgeOldFiles = true
+}
+
+generateLexer {
+    sourceFile = file('src/main/java/com/palantir/gradle/versions/intellij/VersionProps.flex')
+    targetOutputDir = file('src/main/generated/com/palantir/gradle/versions/intellij')
```
CRogers: Same for this.

idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/Contents.java
```diff
@@ -0,0 +1,87 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Value.Immutable
+public abstract class Contents {
+    private static final Logger log = LoggerFactory.getLogger(Contents.class);
+
+    protected abstract String pageContent();
+
+    public static ImmutableContents pageContents(URL pageUrl) {
```
CRogers: You should always return the abstract class/interface type, ie `Contents` rather than `ImmutableContents`. You don't want people to use the `Immutable` version of the type outside of this class.

idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/PackageCompletionContributor.java
```diff
@@ -0,0 +1,55 @@
...
+import com.intellij.codeInsight.completion.CompletionProvider;
+import com.intellij.codeInsight.completion.CompletionResultSet;
+import com.intellij.codeInsight.completion.CompletionType;
+import com.intellij.codeInsight.lookup.LookupElementBuilder;
+import com.intellij.patterns.PlatformPatterns;
+import com.intellij.util.ProcessingContext;
+import com.palantir.gradle.versions.intellij.psi.VersionPropsTypes;
+import java.util.List;
+
+public class PackageCompletionContributor extends CompletionContributor {
```
CRogers: Are `PackageCompletionContributor` and `GroupCompletionContributor` essentially the same at the moment? If so, might as well use the same class for them?

idea-plugin/build.gradle
CRogers: In fact, these two tasks outputting to `src/main/generated` or subfolders means that they would currently clobber each other, meaning Gradle runs them every time.

idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/Contents.java
```diff
@@ -0,0 +1,87 @@
...
+        @Nullable String content = null;
+        ProgressIndicator indicator = ProgressManager.getInstance().getProgressIndicator();
+
+        if (indicator == null) {
+            return ImmutableContents.builder().pageContent("").build();
```
CRogers: In general, using "sentinel" values like this is frowned on. It's better to expose the fact the operation can fail in the type - in this case this method would return an `Optional<Contents>` which would be `Optional.empty()` if it can't be created for some reason.

CRogers: This `Optional<...>` also forces the caller to do something with it and handle the case where it's empty.

idea-plugin/build.gradle
FinlayRJW: Have moved - I still seem to need `sourceSets.main.java.srcDirs 'build/generated/sources/intellij-parser/main/java'` is there a way to remove this?

idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/PackageCompletionContributor.java
FinlayRJW: yep! have combined into a single FolderCompletionContributor

idea-plugin/src/main/generated/com/palantir/gradle/versions/intellij/VersionPropsLexer.java
CRogers, line 1: This has been accidentally checked in I think.

idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/Contents.java
FinlayRJW: Thanks for the tip! make alot of the upstream code simpler for checking if it is empty as well

idea-plugin/src/main/generated/com/palantir/gradle/versions/intellij/VersionPropsLexer.java
FinlayRJW: whoops!

idea-plugin/build.gradle
```diff
@@ -0,0 +1,59 @@
...
+    sourceFile = file('src/main/java/com/palantir/gradle/versions/intellij/VersionProps.bnf')
+    targetRootOutputDir = file('build/generated/sources/intellij-parser/main/java')
+    pathToParser = 'com/palantir/gradle/versions/intellij/parser/VersionPropsParser'
+    pathToPsiRoot = 'com/palantir/gradle/versions/intellij/psi'
+    purgeOldFiles = true
+}
+
+generateLexer {
+    sourceFile = file('src/main/java/com/palantir/gradle/versions/intellij/VersionProps.flex')
+    targetOutputDir = file('build/generated/sources/intellij-parser/main/java/com/palantir/gradle/versions/intellij')
```
CRogers: I think this needs to be in a different source set than the parser one - otherwise Gradle thinks they are going to clobber each other. Maybe in `intellij-lexer` here?

idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/Contents.java
```diff
@@ -0,0 +1,88 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Value.Immutable
+public abstract class Contents {
+    private static final Logger log = LoggerFactory.getLogger(Contents.class);
+
+    protected abstract String pageContent();
+
+    public static Optional<Contents> pageContents(URL pageUrl) {
```
CRogers: This class feels a bit odd being an immutables class. It does complicated mutable stuff (interact with intellij, http request) rather than just shuffling objects, then we extract the string out of it almost immediately. I'd consider just making this a static method on a utility class until we produce a type that has some behaviour (basically instance methods on it).

idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/Contents.java
```diff
@@ -0,0 +1,88 @@
...
+            };
+            Future<String> future = ApplicationManager.getApplication().executeOnPooledThread(task);
+            content = com.intellij.openapi.application.ex.ApplicationUtil.runWithCheckCanceled(future::get, indicator);
+        } catch (InterruptedException | ProcessCanceledException e) {
+            log.debug("Fetch operation was cancelled", e);
+        } catch (Exception e) {
+            log.warn("Failed to fetch contents", e);
+        }
+        return Optional.of(ImmutableContents.builder()
+                .pageContent(Objects.requireNonNullElse(content, ""))
```
CRogers: Another empty string sentinel value - this should probably be an optional again.

idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/Contents.java
```diff
@@ -0,0 +1,88 @@
...
+                connection.disconnect();
+                return result.toString();
+            };
+            Future<String> future = ApplicationManager.getApplication().executeOnPooledThread(task);
+            content = com.intellij.openapi.application.ex.ApplicationUtil.runWithCheckCanceled(future::get, indicator);
+        } catch (InterruptedException | ProcessCanceledException e) {
+            log.debug("Fetch operation was cancelled", e);
+        } catch (Exception e) {
+            log.warn("Failed to fetch contents", e);
+        }
```
CRogers: If you extract out this http stuff to a method, I think we can avoid the `@Nullable String content = null`.

CRogers: Would return an `Optional<String>`

idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/VersionPropsFileType.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @Override
+    public String getDescription() {
+        return "VersionProps language file";
+    }
+
+    @Override
+    public String getDefaultExtension() {
+        return "props";
```
CRogers, line 43: I wonder if results in our plugin being used on all `props` files? Or do we constrain it to working on just `versions.props` somehow.

FinlayRJW: Yes it did, the fix was in `plugin.xml` where `fileType` needed passing `filesNames` rather than an extension

FinlayRJW: Ive also added a test to ensure this is not regressed

idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/Contents.java
FinlayRJW: So I think I've done what you where looking for here - I have converted this to a utility class and split the HTTP stuff into a separate method let me know if it looks good

idea-plugin/src/main/java/com/palantir/gradle/versions/intellij/Contents.java
FinlayRJW: should have done this now

idea-plugin/build.gradle
FinlayRJW: Awesome - have moved

FinlayRJW:
this has been moved into its own repo - see https://github.com/palantir/gradle-consistent-versions-idea-plugin

PR #1235, opened 2024-09-17
Set group and version on all projects

CRogers:
## Before this PR
The previous code to set versions worked fine on a single project setup, but if people were to add another project it would not work.

## After this PR
Make it work for multi-project environment.


felixdesouza:
👍

PR #1245, opened 2024-10-08
Add WstxInputFactory and WstxOutputFactory to XmlMapper

FinlayRJW:
## Before this PR
We were not specifying which Input/Output factories to use for our XMLMapper.
If not specified, the default mapper is initialised using the class loader [here](https://github.com/FasterXML/jackson-dataformat-xml/blob/jackson-dataformat-xml-2.15.2/src/main/java/com/fasterxml/jackson/dataformat/xml/XmlFactory.java#L122-L133).

When not specified, we might be picking a Stax XMLStreamWriter from another library which can then cause failure 

## After this PR
==COMMIT_MSG==
Add WstxInputFactory and WstxOutputFactory to XmlMapper
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-1245.v2.yml
+++ b/changelog/@unreleased/pr-1245.v2.yml
@@ -1,0 +1,5 @@
+type: fix
+fix:
+  description: Add WstxInputFactory and WstxOutputFactory to XmlMapper
+  links:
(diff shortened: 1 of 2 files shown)


felixdesouza:
👍  approved pending sanity check on problem repo

PR #1272, opened 2024-11-19
Require gradle-consitent-versions idea plugin for all projects using GCV

FinlayRJW:
## Before this PR
We had no way of ensuring that the idea plugin was applied for all GCV users

## After this PR
We have duplicated the method used for gradle jdks to ensure the idea plugin is added to the external deps
==COMMIT_MSG==
Require gradle-consitent-versions idea plugin for all projects using GCV
==COMMIT_MSG==

## Possible downsides?
We have duplicated code and should really extract this to a separate plugin to make the management unified


build.gradle
```diff
@@ -49,6 +49,25 @@ allprojects {
...
+    }
+    test {
+        groovy {
+            srcDirs = ['src/test/groovy', 'src/test/java']
+        }
+        java {
+            srcDirs = []
+        }
+    }
+}
```
CRogers: This is not good as it essentially moves all the java code to be compiled by the groovy plugin. Errorprone only runs on java code (so this stops all the errorprones). I'd do what gradle-jdks does and make a subproject project called `gradle-consistent-versions-idea` (the jdks one has one called -groovy) and put this code there under the groovy plugin.

CRogers:
:+1:

FinlayRJW:
👍

PR #1274, opened 2024-11-20
Fix `ConfigureIdeaPluginXml` by removing use of internal `ImmutableMap` class

FinlayRJW:
## Before this PR
Sometimes the `ConfigureIdeaPluginXml. updateIdeaXmlFile` method would fail with `Caused by: java.lang.NoClassDefFoundError: org/gradle/internal/impldep/com/google/common/collect/ImmutableMap` 

This was due to the use of the `ImmutableMap` class which comes from `org.gradle.internal.impldep.com.google.common.collect`  this is internal and not always available at runtime

## After this PR
`ImmutableMap` has been replaced with a standard Groovy map literal which will always be available at runtime
==COMMIT_MSG==
Fix `ConfigureIdeaPluginXml` by removing use of internal `ImmutableMap` class
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-1274.v2.yml
+++ b/changelog/@unreleased/pr-1274.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Fix `ConfigureIdeaPluginXml` by removing use of internal `ImmutableMap`
+    class
+  links:
+  - https://github.com/palantir/gradle-consistent-versions/pull/1274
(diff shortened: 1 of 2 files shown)


felixdesouza:
👍  curious, why have this bit in groovy at all? Checkstyle/Errorprone would have caught the internal class import had it been written in java

FinlayRJW:
> 👍 curious, why have this bit in groovy at all? Checkstyle/Errorprone would have caught the internal class import had it been written in java

For now this is all just copied from `gradle-jdks` so we could release this week - long term we want to pull it all out into a single gradle plugin and I suspect we will drop the groovy

PR #1290, opened 2025-01-28
Groovy based interop API

CRogers:
## Before this PR
In a variety of internal and external projects, we use "groovy interop" classes to change stuff on the GCV extension without actually loading GCV classes.

This is because GCV (for some reason) is the only real project which we decided to excavate out to use the `plugins` block. When a plugin listed in the buildscript tries to access an extension from from a plugin in the plugins block, the classes for the extension are in different classloaders. This means that you get `ClassCastException` (??? need to recheck)

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


CRogers:
This is silly, you can just use `compileOnly` and get actual types and not need another API jar or interop jar.

PR #1322, opened 2025-04-25
Allow `gradle-consistent-version-idea-plugin` idea plugin dependency configuration via `idea-configuration plugin`

FinlayRJW:
## Before this PR
We had our own implementation to configure the external dependancy xml, this is now managed by the shared `idea-configuration` plugin

## After this PR
We use the `idea-configuration` plugin to add the dependancy on `gradle-consitent-version-idea-plugin`

==COMMIT_MSG==
Allow `gradle-consitent-version-idea-plugin` idea plugin dependency configuration via `idea-configuration plugin`
==COMMIT_MSG==


gradle-idea-configuration-interop/src/main/groovy/com/palantir/gradle/versions/IdeaConfigurationInterop.groovy
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.palantir.gradle.versions
+
+import org.gradle.api.Project
+
+class IdeaConfigurationInterop {
```
CRogers: This is an anti-pattern imo and solves a problem I think no longer exists. I'd forgo the groovy and just make a regular java dependency on the plugin jar, always apply the plugin then interact with it through java.

CRogers: Especially if it's a required dep

FinlayRJW: hmm this causes issues with GCV due to java version miss-matches:

```
  bad class file: /Users/finlayw/.gradle/caches/modules-2/files-2.1/com.palantir.gradle.gradleideaconfiguration/gradle-idea-configuration/0.1.0/40af77c55cc565eae29141956b96f26cb3ad3666/gradle-idea-configuration-0.1.0.jar(/com/palantir/gradle/gradleideaconfiguration/IdeaConfigurationExtension.class)
    class file has wrong version 61.0, should be 55.0
    Please remove or make sure it appears in the correct subdirectory of the classpath.
```

Should I drop the java version on gradle idea configuration or bump it in GCV?

changelog/@unreleased/pr-1322.v2.yml
```diff
@@ -0,0 +1,6 @@
+type: improvement
+improvement:
+  description: Allow gradle-consitent-version-idea-plugin idea plugin dependency configuration
+    via idea-configuration plugin
```
felixdesouza:
```suggestion
  description: Allow `gradle-consistent-version-idea-plugin` idea plugin dependency configuration
    via `idea-configuration plugin`
```

src/main/java/com/palantir/gradle/versions/ConsistentVersionsPlugin.java
```diff
@@ -32,7 +34,10 @@ public final void apply(Project project) {
         project.getPluginManager().apply(VersionsPropsPlugin.class);
         project.getPluginManager().apply(GetVersionPlugin.class);
         project.getPluginManager().apply(VersionsPropsIdeaPlugin.class);
-        project.getPluginManager().apply(RequireConsistentVersionsIdeaPlugin.class);
+
+        project.getPluginManager().apply(IdeaConfigurationPlugin.class);
+        IdeaConfigurationExtension extension = project.getExtensions().getByType(IdeaConfigurationExtension.class);
+        extension.externalDependency("gradle-consistent-versions", "0.9.0");
```
felixdesouza, line 40: what's the deal with this version/why is it hard coded?

FinlayRJW: its only used in one place and used to be hard coded anyway - can pull out into a variable if that is neater

felixdesouza: it's more of, what version should it be? should it track the same version as this plugin? if it's hardcoded, are things getting up to date?

FinlayRJW: Historically we have just had all the versions pinned to a minimum I guess we could transition to having them pinned to the plugin version. - this would probably help encourage people to upgrade there plugins as well

felixdesouza (approved):

PR #1352, opened 2025-06-12
make `writeVersionsLock` task configuration cacheable

FinlayRJW:
## Before this PR
`writeVersionsLock` would fail when using configuration cache for 2 reasons:
 - `checkWriteLocksShouldBeRunning` in `WriteVersionsLocksMarkerTask` captured`Project`
 ```
- [error] invocation of `Task.project` at execution time is unsupported.
    - task `:writeVersionsLocks` of type `com.palantir.gradle.versions.WriteVersionsLocksMarkerTask`
        - Exception at `com.palantir.gradle.versions.WriteVersionsLocksMarkerTask.checkWriteLocksShouldBeRunning(WriteVersionsLocksMarkerTask.java:28)`
org.gradle.api.InvalidUserCodeException: Invocation of 'Task.project' by task ':writeVersionsLocks' at execution time is unsupported.
	at org.gradle.api.DefaultTask.getProject(DefaultTask.java:59)
	at com.palantir.gradle.versions.WriteVersionsLocksMarkerTask.checkWriteLocksShouldBeRunning(WriteVersionsLocksMarkerTask.java:28)
	at java.base/java.lang.Thread.run(Thread.java:1583)
```
 - `upToDateWhen(_ignored -> false)` is not configuration cacheable
```
* Exception is:
org.gradle.configurationcache.ConfigurationCacheError: Configuration cache state could not be cached: field `specs` of `org.gradle.api.specs.AndSpec` bean found in task `:writeVersionsLocks` of type `com.palantir.gradle.versions.WriteVersionsLocksMarkerTask`: error writing value of type '[Lorg.gradle.api.specs.Spec;'
	at org.gradle.configurationcache.problems.ProblemsListener$DefaultImpls.onError(ProblemsListener.kt:37)
	at org.gradle.configurationcache.problems.ConfigurationCacheProblems.onError(ConfigurationCacheProblems.kt:37)
	at org.gradle.configurationcache.serialization.AbstractIsolateContext.onError(Contexts.kt:399)
	at org.gradle.configurationcache.serialization.beans.BeanPropertyWriterKt.writeNextProperty(BeanPropertyWriter.kt:100)
	at org.gradle.configurationcache.serialization.beans.BeanPropertyWriter.writeStateOf(BeanPropertyWriter.kt:53)
	at org.gradle.configurationcache.serialization.codecs.BeanCodec.writeBeanOf(BeanCodec.kt:59)
	at org.gradle.configurationcache.serialization.codecs.BeanCodec.encode(BeanCodec.kt:37)
	at org.gradle.configurationcache.serialization.CombinatorsKt$reentrant$1$encodeLoop$1.invokeSuspend(Combinators.kt:143)
```

## After this PR
The above error has been fixed and the tests have been updated to prevent regression
==COMMIT_MSG==
make `writeVersionsLock` task configuration cacheable
==COMMIT_MSG==

- Now `WriteVersionsLocksMarkerTask` take an input `getShouldWriteLocks` which avoids capturing the `Project` object this is then configured when the task is registered
- Switched to `doNotTrackState`

Tests where updated to run with configuration cache this required the version of nebula test to be bumped

### CODE REVIEW

File: src/test/groovy/com/palantir/gradle/versions/IntegrationSpec.groovy [hunk 1]
```diff
@@ -35,4 +35,21 @@ class IntegrationSpec extends IntegrationTestKitSpec {
...
+     */
+    boolean runTasksWithConfigurationCache(String... tasks) {
+        def firstRun = createRunner(tasks + ['--configuration-cache'] as String[]).build()
+        assert firstRun.output.contains('Configuration cache entry stored.'),
+                "Expected first run to store configuration cache, but output was: ${firstRun.output}"
+
+        def secondRun = createRunner(tasks + ['--configuration-cache'] as String[]).build()
+        assert secondRun.output.contains('Configuration cache entry reused.'),
+                "Expected second run to reuse configuration cache, but output was: ${secondRun.output}"
+
```
felixdesouza, line 52: are you going to delete the configuration cache directory like in some of the other PRs?

It would be good to keep the impls the same!

felixdesouza:
It's not clear to me why:

```
upToDateWhen(_ignored -> false)
```

is inherently not configuration cache compatible. I'm sure there are plenty of tasks in plugins you've made cc-compat where we use this method.

My guess is that there's another spec inside that it can't serialise. Are you able to hook up a debugger and have a look?

src/main/java/com/palantir/gradle/versions/WriteVersionsLocksMarkerTask.java
```diff
@@ -18,14 +18,17 @@
...
 import org.gradle.api.DefaultTask;
 import org.gradle.api.GradleException;
+import org.gradle.api.provider.Property;
+import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.TaskAction;
 
-public class WriteVersionsLocksMarkerTask extends DefaultTask {
+public abstract class WriteVersionsLocksMarkerTask extends DefaultTask {
+    @Input
+    public abstract Property<Boolean> getShouldWriteLocks();
```
kelvinou01, line 27: nit: Should we rename this (and `VersionsLocksPlugin::shouldWriteLocks`) to be more descriptive of what it actually is? i.e. `isDirectlyTriggeredByHuman` or something with that meaning?

kelvinou01: nit nit: Oh and also, since we're now using properties by default, I think omitting the `get-` prefix in front of property getter methods is fine and will make the code a bit nicer? For example, 

```java
if (isDirectlyTriggeredByHuman().get())
```

seems more intuitive to me than

```java
getShouldWriteLocks().get()
```

kelvinou01 (changes requested):

FinlayRJW:
> It's not clear to me why:
> 
> ```
> upToDateWhen(_ignored -> false)
> ```
> 
> is inherently not configuration cache compatible. I'm sure there are plenty of tasks in plugins you've made cc-compat where we use this method.
> 
> My guess is that there's another spec inside that it can't serialise. Are you able to hook up a debugger and have a look?

I've done some more digging and its the lambda that is breaking this:

```
Configuration cache state could not be cached: field `specs` of `org.gradle.api.specs.AndSpec` bean found in task `:writeVersionsLocks` of type `com.palantir.gradle.versions.WriteVersionsLocksMarkerTask`: error writing value of type '[Lorg.gradle.api.specs.Spec;'
> Unable to make field private final java.lang.Object[] java.lang.invoke.SerializedLambda.capturedArgs accessible: module java.base does not "opens java.lang.invoke" to unnamed module
```

This is fixed by updating to:
```java
                    writeVersionsLocks.getOutputs().upToDateWhen(new Spec<>() {
                        @Override
                        public boolean isSatisfiedBy(Task _task) {
                            return false;
                        }
                    });
```

But I still prefer the aesthetic of `task.doNotTrackState("Reason")` unless there is a reason we don't like this

src/main/java/com/palantir/gradle/versions/WriteVersionsLocksMarkerTask.java
FinlayRJW: I think as it is annotated with `@Input`  you need to prefix with `get` see:

```
* Causal chain is:
	org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':validatePlugins'.
	org.gradle.internal.execution.WorkValidationException: Plugin validation failed with 1 problem:
  - Error: Type 'com.palantir.gradle.versions.WriteVersionsLocksMarkerTask' function 'isDirectlyTriggeredByHuman()' should not be annotated with: @Input.
    
    Reason: Input/Output annotations are ignored if they are placed on something else than a getter.
    
    Possible solutions:
      1. Remove the annotations.
      2. Rename the method.
```

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -260,7 +260,9 @@ public final void apply(Project project) {
         // we can't just have the task run the write locks code as we need to write the locks in afterEvaluate.
         project.getTasks()
                 .register(WRITE_VERSIONS_LOCKS_TASK, WriteVersionsLocksMarkerTask.class, writeVersionsLocks -> {
-                    writeVersionsLocks.getOutputs().upToDateWhen(_ignored -> false);
+                    writeVersionsLocks.getDirectlyTriggeredByHuman().set(VersionsLockPlugin.shouldWriteLocks(project));
+                    // upToDateWhen(_ignored -> false) is not configuration cacheable use doNotTrackState instead
+                    writeVersionsLocks.doNotTrackState("This task should always run.");
```
CRogers, line 264: Fwiw `doNotTrackState` is apparently more efficient than `upToDateWhen(false)`, because it turns off all state tracking so it doesn't even bother to compare to the current and previous states during up-to-dateness checks. It might be worth adding this to the list of errorprones as an autofix, would be a straightforward one.

CRogers: You can also apparently annotate the task with `@UntrackedTask("reason")` which might be more discoverable.

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -260,7 +260,9 @@ public final void apply(Project project) {
         // we can't just have the task run the write locks code as we need to write the locks in afterEvaluate.
         project.getTasks()
                 .register(WRITE_VERSIONS_LOCKS_TASK, WriteVersionsLocksMarkerTask.class, writeVersionsLocks -> {
-                    writeVersionsLocks.getOutputs().upToDateWhen(_ignored -> false);
+                    writeVersionsLocks.getDirectlyTriggeredByHuman().set(VersionsLockPlugin.shouldWriteLocks(project));
```
CRogers: I think `getShouldWriteLocks` might make more sense here. It's often the not the case that a human is actually triggering this - it's often a robot.

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
FinlayRJW: added :)

CRogers (approved):

PR #1393, opened 2025-07-30
Ensure `--write-locks` works when used in included builds

felixdesouza:
## Before this PR
When you try to use GCV in both the main build and the included build, you either get:
* Gradle 7: incorrect resolution
* Gradle 8: errors where the main build is trying to consume the inner build's internal configurations used to calculate its lock state

```
project : (requested: 'project :' because: requested)
        Failures:
           - Could not resolve project :.
           - Module ':8-8-works-with-included-builds' has been rejected:
     Cannot select module with conflict on capability 'gcv:versions-props:0' also provided by [com.palantir.included-build:included-build:unspecified(gcvVersionsPropsConstraints)]
```

We should not be using the internal configurations of included builds. The way we should be thinking about included builds is to imagine if they were published jars in a separate repo and we were consuming the artifacts that way.

This is happening because Gradle is configuration and variant aware across included builds! (Which is cool, just a bit annoying in this instance).

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

To that end, considering all of the configurations and dependencies we use are annotated with attributes pertaining to GCV. We can add another attribute that uniquely identifies a gradle build within the build tree. 

This leverages [`Project#getBuildTreePath`](https://docs.gradle.org/current/javadoc/org/gradle/api/Project.html#getBuildTreePath()) which gets the unique path of a project within the entire hierarchy. This is added in `Gradle 8.3`, for versions below - `(7.6.4, 8.2)` - we fallback to using `GradleInternal#getIdentityPath().getPath()` which is the same thing (provided we call `getBuildTreePath` on the root project of the build in question.

This means that when GCV is requesting dependencies and configurations, it will narrow down the candidates to be only those produced by the current build instance. This allows `--write-locks` to succeed, and then normal variant selection can occur e.g. get `apiClasses` or `runtimeClasses` variant from the included build which is substituted.

### Utilities
* To make it a bit easier to see what's going on, we extract some sort of `Attribute` action applier which will apply the base set of attributes i.e. `org.gradle.usage=consistent-versions-usage` and `com.palantir.consistent-versions.build-path=<build-tree-path>` and it will apply it consistently to the things we care about.

### DIFF
--- a/src/main/java/com/palantir/gradle/versions/ConsistentVersionsPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/ConsistentVersionsPlugin.java
@@ -42,8 +42,8 @@ public class ConsistentVersionsPlugin implements Plugin<Project> {
         if (!project.getRootProject().equals(project)) {
             throw new GradleException("Must be applied only to root project");
         }
-        project.allprojects(p -> {
-            AttributesSchema attributesSchema = p.getDependencies().getAttributesSchema();
+        project.allprojects(proj -> {
+            AttributesSchema attributesSchema = proj.getDependencies().getAttributesSchema();
             attributesSchema.attribute(GcvBuildPath.ATTRIBUTE);
         });
         project.getPluginManager().apply(VersionsLockPlugin.class);
--- a/src/main/java/com/palantir/gradle/versions/ConsistentVersionsPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/ConsistentVersionsPlugin.java
@@ -19,8 +19,11 @@ package com.palantir.gradle.versions;
 import com.palantir.gradle.ideaconfiguration.IdeaConfigurationExtension;
 import com.palantir.gradle.ideaconfiguration.IdeaConfigurationPlugin;
 import org.gradle.api.GradleException;
+import org.gradle.api.Named;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
+import org.gradle.api.attributes.Attribute;
+import org.gradle.api.attributes.AttributesSchema;
 
 public class ConsistentVersionsPlugin implements Plugin<Project> {
     static final String CONSISTENT_VERSIONS_USAGE = "consistent-versions-usage";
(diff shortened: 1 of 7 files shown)


src/main/java/com/palantir/gradle/versions/ConsistentVersionsPlugin.java
```diff
@@ -45,4 +60,57 @@ public final void apply(Project project) {
             });
         });
     }
+
+    public abstract static class GcvAttributes {
+        @Inject
+        protected abstract Gradle getGradle();
+
+        @Inject
+        protected abstract ObjectFactory getObjects();
```
CRogers, line 69: Injecting stuff really does make the code nicer elsewhere!

src/test/groovy/com/palantir/gradle/versions/ConsistentVersionsPluginIntegrationSpec.groovy
```diff
@@ -452,4 +452,158 @@ class ConsistentVersionsPluginIntegrationSpec extends IntegrationSpec {
...
+
+    def "#gradleVersionNumber: works with included builds"() {
+        setup:
+        gradleVersion = gradleVersionNumber
+
+        // add included build
+        def includedBuild = directory("included-build")
+        settingsFile << """
+            includeBuild '${includedBuild.name}'
+        """
```
CRogers: `.stripIndent(true)` 🙏🏻

CRogers: And elsewhere

CRogers:
Looks good other than that one comment - using the build path as an attribute is a nice fix to this issue

src/test/groovy/com/palantir/gradle/versions/ConsistentVersionsPluginIntegrationSpec.groovy
felixdesouza: oop, missed this

CRogers (approved):

PR #1403, opened 2025-08-20
Document all getVersion variants

pkoenig10:
It would be helpful to know that `getVersion(moduleName)` exists without needing to read the code.


felixdesouza:
👍

PR #1406, opened 2025-08-26
Move java tests in groovy source set to java source set

CRogers:
## Before this PR
While doing #1001, I noticed these Java tests are in the groovy source set.

## After this PR
Move them to the Java source set, so they get errorprones, checkstyle etc.

## Possible downsides?
Loss of history.


felixdesouza:
👍

PR #1420, opened 2025-09-17
Add optional filtering of published dependency constraints to used dependencies only

FinlayRJW:
## Before this PR
Background: https://pl.ntr/2wT

The `gradle-consistent-versions` plugin currently publishes **all** constraints from `versions.lock` in component metadata (`POMs` and Gradle Module Metadata), even for dependencies that aren't actually used by the published component. This causes:

1. **Bloated POMs**: The `<dependencyManagement>` section includes irrelevant dependencies, confusing users who may think these are actual dependencies
2. **Problematic Module Metadata**: Unlike POMs where `dependencyManagement` is mostly ignored, Gradle Module Metadata applies these constraints by default, potentially affecting dependency resolution unnecessarily

## After this PR
This PR introduces an opt-in property `com.palantir.gradle.versions.filterLockFileConstraints` that, when enabled, filters published constraints to only include dependencies in the actual transitive closure of the component being published.

- When `filterLockFileConstraints=true`, the plugin resolves the compile and runtime classpaths to determine which modules are actually used
- Only constraints for these used modules are included in the published metadata
- Local project constraints (`com.palantir.gradle.versions.publishLocalConstraints`) continue to work as before

This filtering capability previously existed (https://github.com/palantir/gradle-consistent-versions/pull/191) but was removed due to performance concerns (https://github.com/palantir/gradle-consistent-versions/pull/504). By making this opt-in, projects can choose between:
- **Default**: Fast builds, but with all constraints published (same as current)
- **Filtered constraints**: Cleaner published metadata, but with the performance cost of resolving configurations during publishing

==COMMIT_MSG==
Add optional filtering of published dependency constraints

Introduces `com.palantir.gradle.versions.filterLockFileConstraints` property to filter published constraints to only include dependencies actually used by the component, reducing metadata bloat while maintaining backwards compatibility as an opt-in feature.
==COMMIT_MSG==

## Possible downsides?
- Performance impact when enabled due to configuration resolution during publishing  
- May increase build times for projects with many modules


FinlayRJW:
Example:

build.gradle
```gradle
dependencies {
    implementation project(':environment-variables')
    implementation gradleApi()
    api project(':providers')

    annotationProcessor 'org.immutables:value'

    compileOnly 'org.immutables:value::annotations'

    testImplementation 'org.assertj:assertj-core'
    testImplementation 'org.junit.jupiter:junit-jupiter'
    testImplementation 'com.netflix.nebula:nebula-test'
}
```


Before:
```json
      "dependencyConstraints": [
        {
          "group": "com.fasterxml.jackson.core",
          "module": "jackson-annotations",
          "version": {
            "requires": "2.19.2"
          },
          "reason": "Computed from com.palantir.consistent-versions' versions.lock in gradle-utils"
        },
        {
          "group": "com.gradle",
          "module": "develocity-gradle-plugin-adapters",
          "version": {
            "requires": "1.2.1"
          },
          "reason": "Computed from com.palantir.consistent-versions' versions.lock in gradle-utils"
        },
        {
          "group": "org.apache.commons",
          "module": "commons-lang3",
          "version": {
            "requires": "3.18.0"
          },
          "reason": "Computed from com.palantir.consistent-versions' versions.lock in gradle-utils"
        },
        {
          "group": "org.immutables",
          "module": "value",
          "version": {
            "requires": "2.11.3"
          },
          "reason": "Computed from com.palantir.consistent-versions' versions.lock in gradle-utils"
        }
      ],
```
After
```json
      "dependencyConstraints": [
        {
          "group": "org.immutables",
          "module": "value",
          "version": {
            "requires": "2.11.3"
          },
          "reason": "Computed from com.palantir.consistent-versions' versions.lock in gradle-utils"
        }
      ],
```

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -959,29 +959,58 @@ private static void configureUsingConstraints(
...
+                    .register("gcvPublishConstraints", conf -> {
+                        conf.setDescription("Publishable constraints from the GCV versions.lock file");
+                        conf.setCanBeResolved(false);
+                        conf.setCanBeConsumed(false);
+                        conf.getDependencyConstraints().addAll(localProjectConstraints);
+                        conf.getDependencyConstraints()
+                                .addAllLater(maybeFilterConstraintsByUsage(subproject, lockFileConstraints));
+                    });
+
             subproject
```
FinlayRJW: Do we want to make the default behaviour not publishing the dep constraints at all?

kelvinou01:
Let's get this in for now, as it would unblock https://github.com/palantir/suppressible-error-prone/issues/178. Libraries which want to publish error-prones can opt-in to filtering + publishing GMM, while taking the performance hit of always resolving runtimeClasspath and compileClasspath

I tried to get filtering to work without the performance penalty described in https://github.com/palantir/gradle-consistent-versions/pull/504, i.e. to pass this test case, but alas. 

```groovy

    def "#gradleVersionNumber: filtering constraints does not trigger resolution of project dependencies during publication"() {
        setup:
        file('gradle.properties') << '''
            com.palantir.gradle.versions.filterLockFileConstraints = true
        '''
        gradleVersion = gradleVersionNumber

        buildFile << """
            allprojects {
                apply plugin: 'java'
            }
        """.stripIndent(true)

        String publish = """
            apply plugin: 'maven-publish'
            group = 'com.palantir.test'
            version = '1.0.0'
            publishing.publications {
                maven(MavenPublication) {
                    from components.java
                }
            }
        """.stripIndent(true)

        addSubproject('api', """
            $publish
            dependencies {
                implementation 'ch.qos.logback:logback-classic:1.2.3'
            }
            
            // Track whether compileClasspath or runtimeClasspath were resolved
            configurations.compileClasspath.incoming.afterResolve {
                def resolvedFile = file("\$buildDir/compileClasspath-resolved")
                resolvedFile.parentFile.mkdirs()  // Create parent directories
                resolvedFile.text = 'resolved'
            }
            configurations.runtimeClasspath.incoming.afterResolve {
                def resolvedFile = file("\$buildDir/runtimeClasspath-resolved")
                resolvedFile.parentFile.mkdirs()  // Create parent directories
                resolvedFile.text = 'resolved'
            }
        """.stripIndent(true))

        addSubproject('service', """
            $publish
            dependencies {
                implementation project(':api')
                implementation 'junit:junit:4.10'
            }
        """.stripIndent(true))

        if (GradleVersion.version(gradleVersionNumber) < GradleVersion.version("6.0")) {
            settingsFile << """
                enableFeaturePreview('GRADLE_METADATA')
            """.stripIndent(true)
        }

        runTasks('--write-locks')

        when:
        runTasks('generateMetadataFileForMavenPublication')

        then: "api's compileClasspath and runtimeClasspath were NOT resolved during publication"
        !Files.exists(Paths.get(projectDir.getPath(), 'api/build/compileClasspath-resolved'))
        !Files.exists(Paths.get(projectDir.getPath(), 'api/build/runtimeClasspath-resolved'))

        and: "but the metadata files were still generated"
        Files.exists(Paths.get(projectDir.getPath(), 'api/build/publications/maven/module.json'))
        Files.exists(Paths.get(projectDir.getPath(), 'service/build/publications/maven/module.json'))

        where:
    }


```

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
kelvinou01: Yep sounds good to me, as the deps dumped into  `<dependencyManagement> ` is noise anyways, so not publishing any dep constraints to GMM is equivalent to the status quo. Looking for a second opinion before I commit this though (as this is not my PR)

src/test/groovy/com/palantir/gradle/versions/VersionsLockPluginIntegrationSpec.groovy
```diff
@@ -959,7 +1067,102 @@ class VersionsLockPluginIntegrationSpec extends IntegrationSpec {
...
+            dependencies {
+                implementation 'junit:junit:4.10'
+            }
+        """.stripIndent(true))
+
+        if (GradleVersion.version(gradleVersionNumber) < GradleVersion.version("6.0")) {
+            settingsFile << """
+                enableFeaturePreview('GRADLE_METADATA')
+            """.stripIndent(true)
+        }
```
felixdesouza, line 1111: this is never going to be true, minimum version is gradle 7.6.4

felixdesouza: ah, this was a copy pasted test

kelvinou01: should i remove all checks for 6.0?

kelvinou01:
We should really test that this doesn't break publishing, e.g. cutting an RC for a simple service and publishing to sandbox

https://github.com/palantir/gradle-guide/blob/814abed80a5a0ae613210db10b4c1e66b861fe4a/guide/testing-and-running-your-plugins-locally.md#running-your-plugin-in-other-repos-on-ci

We should also try to find what perf concerns were in https://github.com/palantir/gradle-consistent-versions/pull/504

PR #1436, opened 2025-10-07
Add virtual platform constraint plugins for automatic version alignment across group modules

FinlayRJW:
## Before this PR

- **Problem:**  
  Our builds do not currently guarantee that all modules within the same group (e.g., `com.palantir:*`) are resolved to the same version when used together—especially when multiple versions are present due to transitive dependencies or downgrades.  
  Previous attempts ([#1423](https://github.com/palantir/gradle-consistent-versions/pull/1423)) to add constraints via Gradle Module Metadata (GMM) improved this but did not fully enforce alignment, particularly during version downgrades.

## After this PR

- **Solution:**  
  Introduces a new "virtual platform" mechanism for strict version alignment of all Java modules within the same group, using a pair of new plugins:

  - **`ConstraintProducerPlugin`**  
    - Applied automatically to the root project by the new settings plugin.
    - For every group with multiple Java modules, injects a synthetic *virtual platform* dependency constraint (e.g., `consistent-versions.external-virtual-platform.com.palantir:_`) into all Java modules in the group.
    - This constraint is published in the POM and Gradle Module Metadata, making alignment information available to consumers.

  - **`ConstraintConsumerSettingsPlugin`** (published as the `com.palantir.externally-consistent-versions` settings plugin)  
    - Applied in the consumer's `settings.gradle`.
    - During dependency resolution, automatically inspects POMs and metadata for the virtual platform constraint.
    - Uses a component metadata rule to assign all modules with the constraint to a virtual platform, ensuring Gradle always aligns them to the same version—even if a consumer requests a lower version with `force` or `strictly`.
    - Works for both buildscript and runtime classpaths.

  - **Tests:**  
    - Comprehensive integration tests verify that version alignment works as expected in both buildscript and runtime configurations, including cases with forced downgrades, transitive dependencies, and malformed POMs.

## Usage
  Add to `settings.gradle`:
  ```groovy
  plugins {
      id 'com.palantir.externally-consistent-versions' version '<current version>'
  }
  ```

## Why Not Use BOMs?

- **No separate BOM artifact needed:** Alignment metadata is published alongside each module.
- **No explicit imports required:** Alignment is enforced automatically for any consumer using the settings plugin.
- **Works even on downgrades:** Strict alignment is enforced even when versions are forced or strictly constrained, which BOMs cannot guarantee.

==COMMIT_MSG==
Add virtual platform constraint plugins to enable automatic version alignment

- Adds `ConstraintProducerPlugin` to inject a virtual platform constraint into all Java modules in a group.
- Adds `ConstraintConsumerSettingsPlugin` to enforce version alignment at dependency resolution time based on published POM.
- Ensures consumers always resolve all modules in a group to the same version, even when downgrades or transitive mismatches occur.
- Includes comprehensive tests and dependency updates to support the new functionality.
==COMMIT_MSG==

## Possible Downsides

- **Added Complexity:** Introduces new plugins and mechanisms that contributors and consumers should be aware of.
- **Settings Plugin Required:** Consumers must explicitly add the settings plugin to benefit from alignment.


src/main/java/com/palantir/gradle/versions/VirtualPlatformSettingsPlugin.java
```diff
@@ -0,0 +1,196 @@
...
+
+        @Inject
+        protected abstract RepositoryResourceAccessor getRepositoryResourceAccessor();
+
+        @Override
+        public final void execute(ComponentMetadataContext context) {
+            ModuleVersionIdentifier id = context.getDetails().getId();
+
+            // Try GMM first
+            String gmmPath = buildGmmPath(id);
```
FinlayRJW: not sure if we even need to care about the GMM could just look at POM?

FinlayRJW: is it true we will always publish a POM? If so we can probably simplify this

FinlayRJW: I think we should be safe to just use the POM as we always publish a POM

src/main/java/com/palantir/gradle/versions/VirtualPlatformSettingsPlugin.java
```diff
@@ -0,0 +1,142 @@
...
+import org.gradle.api.artifacts.ComponentMetadataContext;
+import org.gradle.api.artifacts.ComponentMetadataRule;
+import org.gradle.api.artifacts.ModuleVersionIdentifier;
+import org.gradle.api.artifacts.repositories.RepositoryResourceAccessor;
+import org.gradle.api.initialization.Settings;
+import org.gradle.api.logging.Logger;
+import org.gradle.api.logging.Logging;
+import org.immutables.value.Value;
+
+public class VirtualPlatformSettingsPlugin implements Plugin<Settings> {
```
CRogers: Settings plugins need to be in their own `-settings` jar/project with zero dependencies. Otherwise they will cause the classloader issue to occur.

src/main/java/com/palantir/gradle/versions/ConsistentVersionsPlugin.java
```diff
@@ -49,6 +49,7 @@ public final void apply(Project project) {
         project.getPluginManager().apply(VersionsPropsPlugin.class);
         project.getPluginManager().apply(GetVersionPlugin.class);
         project.getPluginManager().apply(VersionsPropsIdeaPlugin.class);
+        project.getPluginManager().apply(AddVirtualPlatformPlugin.class);
```
CRogers: I'm just wondering what's the best way to roll this out. There's a few approaches:

1. This approach, which forces everyone using `com.palantir.consistent-versions` to opt into adding the external platform. We need a feature flag to roll out.
2. Add the plugin separately to everyone's repo. Does not tie this to the main consistent-versions plugin. Do not need a feature flag.
3. Get the settings plugin that consumes the metadata apply this plugin in every project in the main build.

2 feels strictly better than 1 to me.
I like 3 but it may conflate the production of the virtual platform metadata with the consumption of it - ie it forces a repo to both. I'm trying to think of examples where we _wouldn't_ want this to happen - I'm not sure there are any. Perhaps we should do 3 in the interests of terseness, we can always add a configuration point in an extension for disabling production/consumption. 3 is also more Isolated Projects compatible (apply plugins in other projects is currently banned in Isolated Projects).

external-consistent-versions/src/main/java/com/palantir/gradle/versions/ConstraintConsumerSettingsPlugin.java
```diff
@@ -0,0 +1,142 @@
...
+
+    @Override
+    public final void apply(Settings settings) {
+        settings.getGradle().allprojects(project -> {
+            project.getDependencies().getComponents().all(VirtualPlatformRule.class);
+            project.getBuildscript().getDependencies().getComponents().all(VirtualPlatformRule.class);
+        });
+    }
+
+    public abstract static class VirtualPlatformRule implements ComponentMetadataRule {
```
CRogers, line 64: We should test the performance of this on a large internal repo wrt to dependency resolution times. There is the option to cache these rules, we should see if we need to.

src/main/java/com/palantir/gradle/versions/VirtualPlatformSettingsPlugin.java
```diff
@@ -0,0 +1,142 @@
...
+            String pomPath = buildPomPath(id);
+            getRepositoryResourceAccessor().withResource(pomPath, resource -> {
+                parseMetadata(resource)
+                        .flatMap(Metadata::extractDependencies)
+                        .filter(deps -> hasVirtualPlatform(deps, id.getGroup()))
+                        .ifPresent(_deps -> assignToPlatform(context, id));
+            });
+        }
+
+        private static Optional<Metadata> parseMetadata(InputStream resource) {
```
CRogers: `parsePom`?

external-consistent-versions/src/main/java/com/palantir/gradle/versions/ConstraintConsumerSettingsPlugin.java
```diff
@@ -0,0 +1,142 @@
...
+                            .module()
+                            .filter(AddVirtualPlatformPlugin.VIRTUAL_PLATFORM_NAME::equals)
+                            .isPresent();
+        }
+
+        private static String buildPomPath(ModuleVersionIdentifier id) {
+            String groupPath = id.getGroup().replace('.', '/');
+            return String.format(
+                    "%s/%s/%s/%s-%s.pom", groupPath, id.getName(), id.getVersion(), id.getName(), id.getVersion());
+        }
```
CRogers, line 118: If we could use `project.getDependencies().createArtifactResolutionQuery` this might be a bit nicer in that you wouldn't need to build up this path manually. `DependencyHandler` is listed as `@ServiceScope({Scope.Project.class})` so you _might_ be able to inject it here, there is a lot of mutable stuff on it though so it might be angry - worth a try?

src/main/java/com/palantir/gradle/versions/VirtualPlatformSettingsPlugin.java
```diff
@@ -0,0 +1,142 @@
...
+            try {
+                return Optional.of(XML_MAPPER.readValue(resource, Metadata.class));
+            } catch (IOException e) {
+                log.debug("Failed to parse POM metadata: {}", e.getMessage());
```
CRogers: If this happens we need to explode so we fix the issue. Silently doing nothing means it will never get fixed.

src/test/groovy/com/palantir/gradle/versions/AddVirtualPlatformPluginIntegrationSpec.groovy
```diff
@@ -0,0 +1,114 @@
...
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import spock.lang.Unroll;
+import org.gradle.util.GradleVersion
+
+import static com.palantir.gradle.versions.GradleTestVersions.GRADLE_VERSIONS
+
+
+@Unroll
+class AddVirtualPlatformPluginIntegrationSpec extends IntegrationSpec {
```
CRogers: We should run these tests with configuration cache so we don't immediately produce something non-configuration-cacheable

src/test/groovy/com/palantir/gradle/versions/AddVirtualPlatformPluginIntegrationSpec.groovy
```diff
@@ -0,0 +1,114 @@
...
+        // Create service subprojects
+        addSubproject('service-a', '// Service A')
+        addSubproject('service-b', '// Service B')
+    }
+
+    def "#gradleVersionNumber: published constraints with platform constraint"() {
+        setup:
+        file('gradle.properties') << 'com.palantir.gradle.versions.addVirtualPlatformConstraint = true'
+
+        if (GradleVersion.version(gradleVersionNumber) < GradleVersion.version("6.0")) {
```
CRogers: Gradle 6 is dead. We should remove all support for it/stop testing against it.

CRogers: Only the most recent two major versions matter.

src/test/groovy/com/palantir/gradle/versions/VirtualPlatformSettingsPluginIntegrationSpec.groovy
```diff
@@ -0,0 +1,293 @@
...
+package com.palantir.gradle.versions
+
+import org.gradle.testkit.runner.GradleRunner
+import org.gradle.testkit.runner.TaskOutcome
+import spock.lang.Unroll
+
+import static com.palantir.gradle.versions.GradleTestVersions.GRADLE_VERSIONS
+
+@Unroll
+class VirtualPlatformSettingsPluginIntegrationSpec extends IntegrationSpec {
```
CRogers: Ditto configuration cache

build.gradle
```diff
@@ -128,6 +129,20 @@ gradlePlugin {
...
         }
+        addVirtualPlatformPlugin {
+            id = 'com.palantir.add-virtual-platform-plugin'
+            implementationClass = 'com.palantir.gradle.versions.AddVirtualPlatformPlugin'
+            displayName = 'Plugin to apply virtual platform constraints to the Gradle Module Metadata.'
+            description = displayName
+            tags.set(['versions'])
+        }
+        virtualPlatformSettingsPlugin {
+            id = 'com.palantir.virtual-platform-plugin'
```
CRogers: These plugin ids need some tweaking. Some guidelines:
1. Plugins do not need `plugin` in the name. In fact - we should follow up adding a check to gradle-guide for this (`gradle` should not be in the name either).
2. It should be clear it's from GCV
3. The settings plugin and main build plugins should be similar.

Perhaps these plugins should be `com.palantir.consistent-versions.external-virtual-platforms` for the settings plugin, and `com.palantir.consistent-versions.produce-external-virtual-platform`? It's a bit of a mouthful, but at least it shows where it came from and implies it something to do with dependencies

Another snappier option: `com.palantir.externally-consistent-versions`? Naming is really hard.

If we go with [option 3 from here](https://github.com/palantir/gradle-consistent-versions/pull/1436/files#r2410997955), the non-settings plugin will be "invisible" so matters a bit less

src/main/java/com/palantir/gradle/versions/VirtualPlatformSettingsPlugin.java
CRogers: We should follow up with a check for this in gradle-guide.

src/main/java/com/palantir/gradle/versions/ExternallyConsistentVersionsSettingsPlugin.java
```diff
@@ -0,0 +1,142 @@
...
+import org.gradle.api.logging.Logging;
+import org.immutables.value.Value;
+
+public class VirtualPlatformSettingsPlugin implements Plugin<Settings> {
+
+    private static final Logger log = Logging.getLogger(VirtualPlatformSettingsPlugin.class);
+
+    @Override
+    public final void apply(Settings settings) {
+        settings.getGradle().allprojects(project -> {
```
CRogers: I'm not sure if this is Isolated Projects compatible - but `getGradle().beforeProject()` definitely is.

src/main/java/com/palantir/gradle/versions/ExternallyConsistentVersionsProducerPlugin.java
```diff
@@ -0,0 +1,95 @@
...
+import org.gradle.api.Plugin;
+import org.gradle.api.Project;
+import org.gradle.api.artifacts.Configuration;
+import org.gradle.api.plugins.JavaPlugin;
+
+public class AddVirtualPlatformPlugin implements Plugin<Project> {
+
+    private static final String ADD_VIRTUAL_PLATFORM_CONSTRAINTS_PROPERTY =
+            "com.palantir.gradle.versions.addVirtualPlatformConstraint";
+    static final String VIRTUAL_PLATFORM_NAME = "palantir-virtual-platform";
```
CRogers: Whatever we choose as our identifier here, we should refer to (gradle-)consistent-versions rather than palantir.

src/main/java/com/palantir/gradle/versions/ExternallyConsistentVersionsProducerPlugin.java
```diff
@@ -0,0 +1,95 @@
...
+                .filter(VersionsLockPlugin::isJavaLibrary)
+                .collect(Collectors.groupingBy(project -> String.valueOf(project.getGroup()), Collectors.toSet()))
+                .entrySet()
+                .stream()
+                .filter(entry -> entry.getValue().size() > 1)
+                .forEach(entry -> applyVirtualPlatformPerGroup(entry.getKey(), entry.getValue()));
+    }
+
+    private void applyVirtualPlatformPerGroup(String groupName, Set<Project> projectGroup) {
+        String platformCoordinates = groupName + ":" + VIRTUAL_PLATFORM_NAME;
```
CRogers: I think in the future we may want to have multiple external virtual platforms per repo. The clearest example of this could be a different platform for `-api` jars and non-`-api` jars - I don't think they necessarily need to have the same version. We may _never_ do this, but we should at least keep our options open with how we define this format.

Perhaps we could have the format `consistent-versions.external-virtual-platform-com.maven.group:_` which gets converted to the virtual platform `com.maven.group:_`?

This means if we want to add multiple virtual platforms in the future, we could do this `com.palantir.foo:_` and `com.palantir.foo:_-api`. This would replicate/be the same virtual platform as when people do this in their `versions.props`:

```properties
com.palantir.foo:* = 1.1.1
com.palantir.foo:*-api = 1.2.0
```

We could always make a new "version" of the fake constraint (eg prefixed with `consistent-versions.external-virtual-platform-v2` instead) but this makes us "forwards compatible" with all the power we currently have in `versions.props`.

build.gradle
```diff
@@ -128,6 +129,20 @@ gradlePlugin {
             description = displayName
             tags.set(['versions'])
         }
+        addVirtualPlatformPlugin {
```
CRogers: We definitely need a section in the readme about this, especially why we're doing this compared to the "constraints on other projects" like in boms approach

src/main/java/com/palantir/gradle/versions/ConsistentVersionsPlugin.java
FinlayRJW: I think 3 is the best option, if we opt for 2 we would be adding both plugins to everyones repos anyway and going for 3 will make rollout a lot simpler

src/test/groovy/com/palantir/gradle/versions/AddVirtualPlatformPluginIntegrationSpec.groovy
FinlayRJW: I'm not sure we can the `generatePomFileForMavenPublication` task fails with configuration cache enabled:

```
Configuration cache state could not be cached: field `finalizers` of `org.gradle.internal.xml.XmlTransformer` bean found in field `xmlTransformer` of `org.gradle.api.publish.maven.internal.tasks.MavenPomFileGenerator$MavenPomSpec` bean found in field `value` of `org.gradle.internal.Try$Success` bean found in field `result` of `org.gradle.internal.serialization.Cached$Fixed` bean found in field `mavenPomSpec` of task `:service-a:generatePomFileForMavenPublication` of type `org.gradle.api.publish.maven.tasks.GenerateMavenPom`: error writing value of type 'java.util.ArrayList'
```

And I'm 99% this is unrelated to this plugin

FinlayRJW: can run `runTasksWithConfigurationCache('--write-locks')` so will enable that at the least

external-consistent-versions/src/main/java/com/palantir/gradle/versions/ConstraintConsumerSettingsPlugin.java
FinlayRJW: looks like we can't inject it `> No service of type DependencyHandler available in implicit inputs capturing instantiator services.`

external-consistent-versions/src/main/java/com/palantir/gradle/versions/ConstraintProducerPlugin.java
```diff
@@ -0,0 +1,82 @@
...
+
+    static final String VIRTUAL_PLATFORM_NAME = "palantir-virtual-platform";
+
+    @Override
+    public final void apply(Project rootProject) {
+        if (!rootProject.equals(rootProject.getRootProject())) {
+            throw new IllegalStateException("AddVirtualPlatformPlugin must be applied to the root project");
+        }
+
+        rootProject.afterEvaluate(project -> {
```
crogoz, line 43: is this still configuration cacheable ?

src/main/java/com/palantir/gradle/versions/ExternallyConsistentVersionsProducerPlugin.java
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        rootProject.afterEvaluate(project -> {
+            project.getAllprojects().stream()
+                    .filter(VersionsLockPlugin::isJavaLibrary)
+                    .collect(Collectors.groupingBy(proj -> String.valueOf(proj.getGroup()), Collectors.toSet()))
+                    .entrySet()
+                    .stream()
+                    .filter(entry -> entry.getValue().size() > 1)
+                    .forEach(entry -> applyVirtualPlatformPerGroup(entry.getKey(), entry.getValue()));
```
crogoz: I think you could make this smaller like this - not tested though - 

```suggestion
                    .collect(Collectors.groupingBy(proj -> String.valueOf(proj.getGroup()), Collectors.toSet()))
                    .forEach((groupName, projects) -> {
                        if (projects.size() > 1) {
                            applyVirtualPlatformPerGroup(groupName, projects);
                        }
                    });
```

build.gradle
FinlayRJW: gone with `com.palantir.externally-consistent-versions` and generally renamed everything to match

src/test/groovy/com/palantir/gradle/versions/AddVirtualPlatformPluginIntegrationSpec.groovy
CRogers: Yeah, all the publishing stuff is not configuration cache compatible.

src/main/java/com/palantir/gradle/versions/ExternallyConsistentVersionsProducerPlugin.java
FinlayRJW: Cool I've switched over. I assume we will want to add some kind of extension that lets users create the platforms they want, then some logic in the producer plugin to only add the most specific platform constraint. Or can we add all the platform constraints that match the group and let gradle figure it out.

Say a user adds:

```gradle
externalVirtualPlatforms {
    platform("com.palantir.foo:*") {
        version = "1.1.1"
    }
    platform("com.palantir.foo:*-api")
}
```
say we are `com.palantir.foo:bar-api` would we add both a VP on `com.palantir.foo:_:1.1.1` and `com.palantir.foo:_-api:<latest-version>` or just on the most specific?

external-consistent-versions/src/main/java/com/palantir/gradle/versions/ConstraintProducerPlugin.java
FinlayRJW: I think so - I've added a basic test 

``` groovy
    def "check does not break configuration cache"() {
        expect:
        runTasksWithConfigurationCache('--write-locks')
        runTasksWithConfigurationCache('build')
    }

```

Which I think would blow up if there where any configuration cache issues

src/main/java/com/palantir/gradle/versions/VirtualPlatformSettingsPlugin.java
FinlayRJW: Have moved into its own project

external-consistent-versions/src/main/java/com/palantir/gradle/versions/ConstraintConsumerSettingsPlugin.java
FinlayRJW: Testing on an internal project seems to indicate that on CI no more network requests are made when resolving the `runtimeClasspath` and `compileClasspath` see discussion here https://pl.ntr/2xp

PR #1443, opened 2025-10-16
Support Gradle 9!

felixdesouza:
## Before this PR
GCV did not support Gradle 9 due to the following:

### `ProjectDependency#getDependencyProject` was [deprecated in 8.11](https://github.com/gradle/gradle/commit/09127458b641ceb03ea11fcf4a9456ddb1040e77) and [removed in 9.0](https://github.com/gradle/gradle/commit/3a27d546a713568343413f622c872b2f052ea757)
The fix is to use `ProjectDependency#getPath` where possible.  This worked for a number of usages where we just wanted to display the project path in some string.

Unfortunately, a core part of GCV is looking within project's and inspecting their dependencies and doing so in an `afterEvaluate`. This will inevitably fail when we need to support isolated projects. However, considering we *just* want to support Gradle 9, we can just replace calls to `ProjectDependency#getDependencyProject` with:

```java
getGradle().getRootProject().path(projectDependency#getPath());
```

`ProjectDependency#getPath` was added in `8.11` so for everything up to that, we can just get the path from the `Project` via `Project#getPath` and we can get that off of `ProjectDependency#getProjectDependency`:

```java
projectDependency().getProjectDependency().getPath();
```

#### `nebula.dependency-recommender` forces choice between Gradle 7 and Gradle 9 support

Because of the above removal of `ProjectDependency#getProjectDependency`, `nebula.dependency-recommender` on version `11.0.0` fails on Gradle 9 because of a `MethodNotFoundError` trying [to access](https://github.com/nebula-plugins/nebula-dependency-recommender-plugin/blob/v11.0.0/src/main/groovy/netflix/nebula/dependency/recommender/DependencyRecommendationsPlugin.java#L173) `ProjectDependency#getProjectDependency`.

This was fixed in `nebula.dependency-recommender` version [13.0.0](https://github.com/nebula-plugins/nebula-dependency-recommender-plugin/blob/9013cb878ddfd8b2aca2f462a6a7a961fd48215a/src/main/groovy/netflix/nebula/dependency/recommender/DependencyRecommendationsPlugin.java#L173-L176). However, `ProjectDependency#getPath` was only introduced in Gradle 8.11 meaning that if we upgraded our dependency on the recommender, Gradle 7 builds would fail.

Fortunately, we only had this dependency to ease the migration which has long since completed for most of our internal builds. 

> [!WARNING]
> Now, if people are using the recommender (which they shouldn't?), their builds will not fail if they are using the `RecommendationStrategies#OverrideTransitives`.

### Copied configurations cannot have their role changed once created
After we've wired up the locked configurations, we have to recursively copy those configurations so we don't "freeze" them once we resolve the `unifiedClasspath`. Previously, we were copying the configuration and then marking it as consumable amongst a bunch of other things that allow GCV to figure out all the dependencies.

Gradle has been shifting to being strict about `Configuration` roles i.e. dependency scope, resolvable and consumable. 

Given that the locked configurations are necessarily resolvable, the copied configuration *will* also be resolvable.

To fix this, we still make a recursive copy, but we make a new **consumable** configuration and extend the copy.

Note, there won't be any direct dependencies/constraints on this consumable configuration, so you have to use the `getAllDependencies` or any of the `getAll*` variants. This was not a problem before because `copyRecursive` effectively flattens dependencies and constraints from parent configurations and sticks it directly into the new configuration.

Even when this was fixed, it was not possible to [create a new configuration manually and then add it to the container](https://docs.gradle.org/current/userguide/upgrading_version_8.html#adding_to_configuration_container), as you have to go through the either these paths:
  * `ConfigurationContainer#create` (legacy)
  * `ConfigurationContainer#register` (legacy)
  * `ConfigurationContainer#dependencyScope`
  * `ConfigurationContainer#resolvable`
  * `ConfigurationContainer#consumable`

That said, adding to the configuration container is required when you're running on Gradle 7, so there's a workaround to deal with that.

### Presence of consumable configurations freeze attributes/capabilities and more when any resolution occurs
When gradle wants to do a resolution, and you have a project dependency, it will need to enumerate the variants to consider what will participate in variant matching. Once it has discovered these variants, it will freeze all the different properties about it e.g. attributes, capabilities, hierarchy, excludes etc. This includes variants that won't even be selected in this particular resolution! 

Up until gradle 9, that also included dependencies and constraints, but that was [relaxed in Gradle 9](https://github.com/gradle/gradle/blob/b201878f2ca2e8f8150312a9234473bfe534c908/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/configurations/DefaultConfiguration.java#L1262-L1286). If we were to run on strict mode in Gradle 8, we'd see errors about mutating the configuration, *even* with this PR.

The corollary of this means that we need to set up the metadata i.e. hierarchy, attributes and capabilities before we do the `unifiedClasspath` resolution. We can always add dependencies and constraints to the configurations that weren't observed at a later point.

Note: this somewhat turns into a whack-a-mole exercise where anything that creates a consumable configuration needs to not extend or reference another configuration that will participate in dependency locking. 

For example, `apiElements` is a `consumable` configuration, we inject publishable constraints at some point, but first we have it extend an empty `dependencyScope` configuration named `gcvPublishConstraints`. After we've finished writing locks i.e. done a resolution, we can then inject them into `gcvPublishConstraints` and everything is fine.

### sls-packaging is creating configurations at task execution/input checking time
We have a test `SlsPackagingCompatibilityIntegrationSpec` which checks that sls-packaging works correctly. In the previous version of sls-packaging 7.25.0, Gradle 8 was correctly warning that we had ambiguous transform paths. This is now an error in Gradle 9. See [deprecation notice](https://docs.gradle.org/current/userguide/upgrading_version_8.html#deprecated_ambiguous_transformation_chains). Fortunately, this was fixed (inadvertently?) in sls-packaging [7.31.0](https://github.com/palantir/sls-packaging/releases/tag/7.31.0).

Bumping to 7.31.0 fixes the above error but now, this exposes some other weird behaviour in sls-packaging:

```
Could not determine the dependencies of task ':service:resolveProductDependencies'.
> Could not resolve all dependencies for configuration ':service:runtimeClasspathForProductDependencies'.
   > Could not resolve project :service.
     Required by:
         project ':service'
      > A dependency was declared on configuration 'runtimeClasspathForProductDependencies' of 'project :service' but no variant with that configuration name exists.
```

This is because sls-packaging [creates a configuration](https://github.com/palantir/sls-packaging/blob/54b7f034fc8a7c549797e5feb414866995ab12f5/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/artifacts/DependencyDiscovery.java#L45) as part of calculating [task inputs for the `ResolveProductDependencies` task](https://github.com/palantir/sls-packaging/blob/288362a3d53863297f1358fc149606cf0b426430/gradle-sls-packaging/src/main/java/com/palantir/gradle/dist/pdeps/ProductDependencies.java#L78) which happens way too late. IMO this is bad behaviour for sls-packaging and feels correct to be an error here.

The timeline of things I think is happening in Gradle 9 is as follows:
* we do a resolution of `unifiedClasspath`
* gradle looks [for the candidates for graph variant selection](https://github.com/gradle/gradle/blob/master/platforms/software/dependency-management/src/main/java/org/gradle/internal/component/model/GraphVariantSelector.java#L98) within a project
* gradle looks [for all consumable variants for a project](https://github.com/gradle/gradle/blob/v9.1.0/platforms/software/dependency-management/src/main/java/org/gradle/internal/component/local/model/DefaultLocalComponentGraphResolveState.java#L108)
* [freezes and caches this conversion](https://github.com/gradle/gradle/blob/v9.1.0/platforms/software/dependency-management/src/main/java/org/gradle/internal/component/local/model/DefaultLocalComponentGraphResolveState.java#L98) of `Configuration` to internal format
* does not look at the `Configuration`s in `ConfigurationContainer` again i.e. [this](https://github.com/gradle/gradle/blob/v9.1.0/platforms/software/dependency-management/src/main/java/org/gradle/internal/component/local/model/DefaultLocalComponentGraphResolveState.java#L102) is never run again.

For now, I will disable the this particular test for Gradle 9 pending fixes in sls-packaging to be gradle 9 compatible. I think this is reasonable considering this is not GCV related or better still, GCV forces a resolution earlier than anything else, but what's to say that if GCV was not being used, something else wouldn't trigger dependency resolution earlier than the task inputs realised.

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

* Introduce a `ProjectDependencyWorkaround` class that contains the behaviour changes for the different versions.
  * I'm not super keen on it, because there's no nice way to delete code once you've moved past the version without manually going through it yourself. I have ideas on how to make this better, but somewhat out of scope for this PR.

* I've also removed a bunch of old code that was a workaround for versions of gradle < 5.3!! 
* We weren't handling the subproject names correctly inside `VersionsLockPluginIntegrationSpec` and we were referring to projects that didn't have a corresponding folder which is [now an error](https://docs.gradle.org/current/userguide/upgrading_major_version_9.html#project_directories_must_exist_and_be_writeable) in Gradle 9


src/test/groovy/com/palantir/gradle/versions/VersionsLockPluginIntegrationSpec.groovy
```diff
@@ -356,6 +279,17 @@ class VersionsLockPluginIntegrationSpec extends IntegrationSpec {
         gradleVersionNumber << GRADLE_VERSIONS
     }
 
+    @Override
+    File addSubproject(String name) {
```
felixdesouza, line 283: What we really need to do is to be able to create a subproject from a nested directory, but that's not currently possible from what I could see. This is the next best thing i.e. given `foo:a` create projects `foo` and `foo:a` and then return the latter.

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -374,6 +378,19 @@ private static Map<Project, LockedConfigurations> wireUpLockedConfigurationsByPr
                     subproject,
                     subproject.getConfigurations().getByName(CONSISTENT_VERSIONS_TEST),
                     lockedConfigurations.testConfigurations());
+
+            @SuppressWarnings("for-rollout:ConfigurationAvoidanceRegistration")
```
CRogers: Is this just because it's only recently that `.extendsFrom` takes a `Provider<Configuration>` and we still need support old Gradle? If so, it might still be better to make this a `register` than call `.get()`, and it will be easier for an errorprone to clean up later.

felixdesouza: This was just moved from below, so hadn't even thought about it. The `extendsFrom` is actually just in Gradle 9, so we can't even compile against it yet.

But yeah, can do the register and get!

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -1163,4 +1167,43 @@ public static boolean shouldWriteLocks(Project project) {
         return startParameter.isWriteDependencyLocks()
                 || WRITE_VERSIONS_LOCKS_TASK_NAME_MATCHER.matchesAny(startParameter.getTaskNames());
     }
+
+    @SuppressWarnings("deprecation")
+    public abstract static class ProjectDependencyWorkarounds {
```
CRogers: I wonder if we're going to need to promote this to a utils package eventually - but fine just here for now.

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -1163,4 +1167,43 @@ public static boolean shouldWriteLocks(Project project) {
...
+            GradleVersion removedVersion = GradleVersion.version("9.0");
+            if (GradleVersion.current().compareTo(removedVersion) < 0) {
+                return projectDependency.getDependencyProject();
+            }
+
+            return getGradle().getRootProject().project(projectDependency.getPath());
+        }
+
+        @Inject
+        public ProjectDependencyWorkarounds() {}
```
CRogers, line 1206: Do you need this? I though it was fine with the default constructor?

CRogers: Similarly I thought you could at least make the class/ctor package-private rather than public?

felixdesouza: So, when generating the instance of the class, gradle [uses](https://github.com/gradle/gradle/blob/v9.1.0/platforms/core-configuration/model-core/src/main/java/org/gradle/internal/instantiation/generator/AbstractClassGenerator.java#L222) `Class#getConstructors` instead of `Class#getDeclaredConstructors` and the former only gets *public* constructors.

I've made the class package-private. I initially tried `protected` and then went one step further.

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -635,11 +659,47 @@ private void recursivelyCopyProjectDependenciesWithScope(
...
+
+            Configuration copiedConf = projectDep
+                    .getConfigurations()
+                    .register(targetConf.getName() + "GcvConsumable", conf -> {
+                        conf.setDescription(String.format(
+                                "Copy of the '%s' configuration that can be resolved by"
+                                        + " com.palantir.consistent-versions without resolving the '%s' configuration"
+                                        + " itself.",
+                                targetConf.getName(), targetConf.getName()));
+                        // what will I need to do here?
```
CRogers: ??

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
```diff
@@ -635,11 +659,47 @@ private void recursivelyCopyProjectDependenciesWithScope(
...
+                        conf.setCanBeResolved(false);
+                        // Must set this because we depend on this configuration when resolving unifiedClasspath.
+                        conf.setCanBeConsumed(true); // need to pick one, this is not great
```
CRogers: Mixing comments before the line and after the line is always a bit confusing to me. I prefer before the line comments.

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
felixdesouza: okay, if you have package private class, checkstyle will complain at you for a redundant public modifier. So I think just need to settle on `protected` class with `public` constructor.

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
felixdesouza: fortunately I can delete those comments, they were largely for myself

src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java
felixdesouza: deleting, sorry

CRogers:
:+1:

PR #1520, opened 2025-12-03
Fix GcvConsumable configuration creation for Gradle 9+ compatibility

xjetry:
## Summary
- Adds `ensureBaseGcvConsumable` method to pre-create GcvConsumable configurations during Java plugin setup
- Updates `recursivelyCopyProjectDependenciesWithScope` to reuse existing configurations and update their `extendsFrom` relationship
- Makes configurations visible in Gradle 9+ since `setVisible(false)` prevents variant selection discovery in newer Gradle versions

## Changes
The early configuration creation ensures that variant-based project dependencies can properly resolve to GcvConsumable configurations via Gradle's variant selection mechanism. This is necessary because:

1. In Gradle 9+, `setVisible(false)` prevents configurations from being discovered by variant selection from other projects
2. Inherited project dependencies (like `api(project(":foo"))`) need to resolve to GcvConsumable configurations via variant selection
3. Pre-creating configurations during plugin setup ensures they're available before `afterEvaluate` when dependencies are processed

### DIFF
--- a/changelog/@unreleased/pr-1520.v2.yml
+++ b/changelog/@unreleased/pr-1520.v2.yml
@@ -1,0 +1,7 @@
+type: fix
+fix:
+  description: Fix GcvConsumable configuration creation for Gradle 9+ compatibility
+    by pre-creating configurations during plugin setup and making them visible for
+    variant selection discovery.
+  links:
+  - https://github.com/palantir/gradle-consistent-versions/pull/1520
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/gradle-consistent-versions](https://github.com/palantir/gradle-consistent-versions), @xJetRy! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #1565, opened 2026-01-30
Gradle 9 compatibility: fix unsafe configuration resolution

FinlayRJW:
## Before this PR

In Gradle 9, the `checkUnusedConstraints` task fails with:
```
Resolution of the configuration ':repo:errorproneJavac' was attempted without an exclusive lock. 
This is unsafe and not allowed.
```

The task resolves configurations during input evaluation (when `@Input` properties are fingerprinted), which Gradle 9 strictly disallows. Additionally, the task fails when `configuration-on-demand` mode is enabled.

## After this PR

The `checkUnusedConstraints` task works correctly with Gradle 9 and naturally handles `configuration-on-demand` mode.

The implementation now uses a distributed artifact-based approach:
- Each subproject independently produces a JSON artifact containing its resolved module identifiers during the proper Gradle lifecycle phase
- The root project collects these artifacts through Gradle's variant-aware dependency resolution
- The task performs pure file processing with no configuration resolution
- `configuration-on-demand` works seamlessly since each project independently produces its artifact when configured

==COMMIT_MSG==
Fix Gradle 9 compatibility for `checkUnusedConstraints`

Redesigns the task to use artifact-based data collection instead of
direct configuration resolution during task input evaluation.

Introduces `CheckUnusedConstraintsPlugin` and 
`CheckUnusedConstraintsProjectPlugin` to coordinate data collection
across all subprojects. Each project produces a JSON artifact with
its resolved modules, which the root task aggregates.

This eliminates the "attempted without an exclusive lock" error and
naturally handles configuration-on-demand mode.
==COMMIT_MSG==

## Possible downsides?

- The new approach adds intermediate JSON files to each subproject's `build` directory (`build/tmp/check-unused-constraints/resolved-module-identifiers.json`), though these are temporary build artifacts
- Slightly more complex plugin architecture with two plugin classes instead of inline task configuration
- Build output may show additional configuration tasks being executed in parallel builds


src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
```diff
@@ -79,16 +79,26 @@ public final void apply(Project project) {
...
-                                    .flatMap(proj -> CheckUnusedConstraintsTask.getResolvedModuleIdentifiers(
-                                            proj,
-                                            project.getExtensions().getByType(VersionRecommendationsExtension.class)))
-                                    .collect(Collectors.toSet())));
                         }
                         task.getPropsFile()
                                 .set(project.getLayout().getProjectDirectory().file("versions.props"));
                     });
+
+            // Gradle 9 compatibility: configure the classpath after task graph is ready
```
felixdesouza: this feels `afterEvaluate` adjacent, we should see if there's something better we can do here with a `FileCollection` or some sort of `Buildable` thing.

FinlayRJW: from https://github.com/gradle/gradle/issues/33938#issuecomment-2985249363 I've tried to create a configuration that aggregates all dependencies from all subprojects

FinlayRJW: hmm seems to have broken something will take a deeper dive tomorrow

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsTask.java
```diff
@@ -50,65 +48,31 @@
...
-    final void setPropsFile(File propsFile) {
-        this.propsFileProperty.set(propsFile);
-    }
-
-    @Input
-    public final SetProperty<String> getClasspath() {
-        return classpath;
-    }
+    @Internal
+    public abstract Property<Configuration> getAggregatedConfiguration();
```
felixdesouza: we should not be putting `Configuration` objects inside task objects, for configuration cache compatibility right? please fix

felixdesouza: since you've also marked it internal, you've changed the inputs of this task!

src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
```diff
@@ -73,21 +72,36 @@ public final void apply(Project project) {
...
+            NamedDomainObjectProvider<Configuration> gcvAllDependencies = project.getConfigurations()
+                    .register("gcvAllDependencies", conf -> {
+                        conf.setVisible(false);
+                        conf.setCanBeResolved(true);
+                        conf.setCanBeConsumed(false);
+                        conf.setDescription(
+                                "Aggregated dependencies from all projects for checkUnusedConstraints task");
+                    });
+
+            project.allprojects(subproject -> {
```
felixdesouza: this is no longer equivalent. In the previous code, you have 

* for ***all*** projects, and ***all*** their _resolvable_ configurations modulo some filtering, we find the classpath and union them together.

When you depend on another project, the project being depended upon needs to decide what resolvable configuration to give back to you. This is how you get `apiElements`, `runtimeElements` for example. There's the whole thing about variant selection etc. 

So with what you've done, you now have:

* for *all* projects, resolve ***one*** configuration, and then use that in your calculation.

So in theory you can have resolvable configurations that will no longer be checked, hence this is incorrect.

felixdesouza: it is clearer if you actually just inline `CheckUnusedConstraintsTask.getResolvedModuleIdentifiers`. You can see what you're actually iterating over. (For education purposes that is).

FinlayRJW: sorry this PR should really have been a draft - still trying to get my head around GCV and how best todo this

FinlayRJW: So I think what I want is essentially a unifiedClassPath but includes all configurations not just java ones - if that makes any sense!

felixdesouza: that is one way of doing that! Having another `unifiedClasspath` is probably too strict a requirement here. You only want the module names, right?

FinlayRJW: > You only want the module names, right

Yeah I think so

FinlayRJW: you can see my concept here - https://github.com/palantir/gradle-consistent-versions/pull/1565/changes#diff-8f6332e4bf665c9817bd219407e5081faf2bd8e9b94d16b1e11d14bb2205f4e2R542-R553 I think you could pull this out of `VersionsLockPlugin` if this is the right direction to go

FinlayRJW: hmm I think I need to go the isolated projects route

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
```diff
@@ -0,0 +1,86 @@
...
+            outgoing.setVisible(false);
+            outgoing.attributes(attrs -> {
+                attrs.attribute(Usage.USAGE_ATTRIBUTE, project.getObjects().named(Usage.class, OUTGOING_USAGE));
+            });
+
+            outgoing.getOutgoing()
+                    .artifact(getLayout()
+                            .getBuildDirectory()
+                            .file("check-unused-constraints/resolved-module-identifiers.json")
+                            .map(regFile -> {
```
FinlayRJW, line 73: I tried to shift this into a task by passing the `Provider<Set<ResolvedModule>>` into the task and then just writing them out - but this caused the `Resolution of the configuration ':repo:blah' was attempted without an exclusive lock. This is unsafe and not allowed.` error again

FinlayRJW: `checkUnusedConstraints_with_root_config_depending_on_subproject_and_parallel` demonstrates the type of issue I hit when trying to use a task - I think what we have is fine and seems to work

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
```diff
@@ -0,0 +1,107 @@
...
+            outgoing.getOutgoing()
+                    .artifact(getLayout()
+                            .getBuildDirectory()
+                            .file("tmp/check-unused-constraints/resolved-module-identifiers.json")
+                            .map(regFile -> {
+                                regFile.getAsFile().getParentFile().mkdirs();
+
+                                Set<ResolvedModule> identifiers =
+                                        GradleConfigurations.getResolvableConfigurations(project).stream()
+                                                .filter(conf -> GCV_INTERNAL_PREFIXES.stream()
```
FinlayRJW, line 78: I had some odd issues on gradle 7 only where I needed to filter out the GCV configurations else I got a binary store problem: 

```
* What went wrong:
> Error while evaluating property 'currentLockState' of task ':verifyLocks'.
   > Failed to calculate the value of task ':verifyLocks' property 'currentLockState'.
      > Could not resolve all dependencies for configuration ':unifiedClasspath'.
         > Problems reading data from Binary store in /Volumes/git/gradle-consistent-versions/build/tmp/test/work/.gradle-test-kit/.tmp/gradle12660835282155363298.bin offset 0 exists? true

```

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
```diff
@@ -0,0 +1,107 @@
...
+        ResolutionResult resolutionResult = configuration.getIncoming().getResolutionResult();
+        try {
+            return resolutionResult.getAllComponents().stream()
+                    .map(ResolvedComponentResult::getId)
+                    .filter(cid -> !cid.equals(resolutionResult.getRoot().getId()))
+                    .filter(ModuleComponentIdentifier.class::isInstance)
+                    .map(ModuleComponentIdentifier.class::cast)
+                    .map(mcid -> ResolvedModule.of(configuration.getName(), mcid.getGroup() + ":" + mcid.getModule()));
+        } catch (Exception e) {
+            return Stream.empty();
```
FinlayRJW, line 104: I saw an issue only on gradle 7 where these would sometimes fail with:

```
* What went wrong:
Could not determine the dependencies of task ':checkUnusedConstraints'.
> Could not resolve all dependencies for configuration ':collectedCheckUnusedConstraintsOutgoing'.
   > Value for configuration ':check-unused-constraints-outgoing' artifacts has not been calculated yet.
```

src/test/java/com/palantir/gradle/versions/CheckUnusedConstraintIntegrationTest.java
```diff
@@ -181,6 +180,122 @@ void unused_check_should_use_exact_matching(GradleInvoker gradle, RootProject ro
...
+
+    /**
+     * This test reproduces the scenario where a root-level resolvable configuration has lazy dependency
+     * constraints that resolve a subproject's configuration. When running with --parallel, this causes
+     * "Resolution of the configuration was attempted without an exclusive lock" errors.
+     */
+    @Test
+    @AdditionallyRunWithGradle("9.3.0")
+    void checkUnusedConstraints_with_root_config_depending_on_subproject_and_parallel(
+            GradleInvoker gradle, RootProject rootProject, SubProject serviceA) {
```
FinlayRJW, line 227: This demonstrated a pattern I found some of our internal plugins do that breaks check unused constraints when trying to use a `Task` to save the `ResolvedModule` not sure if the real issue here is in those plugins and we should go that route

FinlayRJW:
closed in favour of https://github.com/palantir/gradle-consistent-versions/pull/1573

PR #1573, opened 2026-02-05
Fix `checkUnusedConstraints` for Gradle 9 and configuration-on-demand

FinlayRJW:
## Before this PR

The `checkUnusedConstraints` task resolved configurations during task input evaluation time. This caused two problems:

1. **Gradle 9 incompatibility**: In Gradle 9, resolving configurations during task input fingerprinting is not allowed without proper locking, causing 'Resolution of configuration was attempted without an exclusive lock' errors.

2. **Configuration-on-demand incompatibility**: When using `--configure-on-demand`, not all projects would be configured, so the task couldn't accurately check all constraints. The workaround was to fail with a message asking users to disable configuration-on-demand.

## After this PR

Refactors `checkUnusedConstraints` to use Gradle's artifact/task dependency model:

- **`CheckUnusedConstraintsProjectPlugin`**: Applied to every project. Registers a `WriteResolvedCoordinatesTask` that resolves the project's configurations and writes module identifiers to a JSON file. Exposes the JSON via a consumable configuration with a custom `Usage` attribute.

- **`CheckUnusedConstraintsPlugin`**: Applied to the root project. Collects all subproject JSON artifacts via a resolvable configuration and passes them to `CheckUnusedConstraintsTask`.

- **Task ordering**: Each project's `WriteResolvedCoordinatesTask` declares its resolvable configurations (e.g. `compileClasspath`, `runtimeClasspath`) as lenient `@InputFiles`. Gradle uses the build dependency metadata from these configurations to order tasks across projects, preventing parallel resolution lock errors.

This approach:
- Resolves configurations at task execution time (not input evaluation time), fixing Gradle 9 compatibility
- Works correctly with configuration-on-demand since task dependencies ensure proper ordering
- Ensures configuration cache compatibility

## Test plan

- Existing tests pass
- Added new test case `checkUnusedConstraints_with_root_config_depending_on_subproject_and_parallel` that reproduces the scenario where a resolvable configuration has lazy dependency constraints that resolve another project's configuration
- Tested and working when used on large project with inter-project deps


src/test/groovy/com/palantir/gradle/versions/ConfigurationOnDemandSpec.groovy
```diff
@@ -384,7 +381,6 @@ class ConfigurationOnDemandSpec extends IntegrationSpec {
         runTasks('build')
 
         then:
-        runTasksAndFail(':checkUnusedConstraints')
```
FinlayRJW, line 387: `:checkUnusedConstraints` now succeeds so don't need to test

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
```diff
@@ -0,0 +1,167 @@
...
+                                    .getBuildDirectory()
+                                    .file("tmp/check-unused-constraints/resolved-module-identifiers.json"));
+
+                    task.getResolvedModules()
+                            .set(getProviderFactory()
+                                    .provider(() -> GradleConfigurations.getResolvableConfigurations(project).stream()
+                                            .flatMap(CheckUnusedConstraintsProjectPlugin::resolvedModules)
+                                            .collect(Collectors.toSet())));
+
+                    task.getDependentProjectModule()
```
FinlayRJW: hmm this doesnt prevent cycles if say you have a util subproject that depends via implementation of a different subproject - then that subproject depends on the utils via a testImplementation

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsPlugin.java
```diff
@@ -0,0 +1,103 @@
...
+        if (rootProject != rootProject.getRootProject()) {
+            throw new GradleException("The CheckUnusedConstraintsPlugin plugin must be applied on the root project");
+        }
+
+        rootProject.allprojects(subproject -> {
+            subproject.getPluginManager().apply(CheckUnusedConstraintsProjectPlugin.class);
+        });
+
+        NamedDomainObjectProvider<Configuration> collectedConfiguration = getConfigurations()
+                .register("collectedCheckUnusedConstraintsOutgoing", collected -> {
```
felixdesouza: this is not an outgoing configuration. This is the resolvable configuration from the root project, it is the consumer.

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
```diff
@@ -0,0 +1,167 @@
...
+import org.gradle.api.attributes.Usage;
+import org.gradle.api.file.ProjectLayout;
+import org.gradle.api.model.ObjectFactory;
+import org.gradle.api.provider.ProviderFactory;
+import org.gradle.api.tasks.TaskProvider;
+import org.gradle.util.GradleVersion;
+
+public abstract class CheckUnusedConstraintsProjectPlugin implements Plugin<Project> {
+
+    static final String OUTGOING_USAGE = "checkUnusedConstraintsOutgoingModuleIdentifiers";
```
felixdesouza: just call this `gcv-check-unused-constraints`

since the producer and consumer need to use the same attributes, it doesn't make sense to call it something for one side of the exchange

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
```diff
@@ -0,0 +1,167 @@
...
+                                .forEach(path -> deps.add(getDependencyHandler().project(Map.of("path", path))));
+                    });
+                });
+
+        TaskProvider<WriteResolvedModulesTask> writeResolvedModulesTask = project.getTasks()
+                .register("writeResolvedModulesTask", WriteResolvedModulesTask.class, task -> {
+                    task.getOutputFile()
+                            .set(getLayout()
+                                    .getBuildDirectory()
+                                    .file("tmp/check-unused-constraints/resolved-module-identifiers.json"));
```
felixdesouza: if you're going to use the `tmp` directory, you should just use `task.getTemporaryDirectory` instead.

src/main/java/com/palantir/gradle/versions/WriteResolvedModulesTask.java
```diff
@@ -0,0 +1,63 @@
...
+     */
+    @InputFiles
+    @PathSensitive(PathSensitivity.NONE)
+    public abstract ConfigurableFileCollection getDependentProjectModule();
+
+    @TaskAction
+    public final void writeResolvedModules() {
+        try {
+            OBJECT_MAPPER.writeValue(
+                    getOutputFile().get().getAsFile(), getResolvedModules().get());
```
felixdesouza: you should write this out in sorted order

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
```diff
@@ -0,0 +1,167 @@
...
+
+    private static Stream<ResolvedModule> resolvedModules(Configuration configuration) {
+        ResolutionResult resolutionResult = configuration.getIncoming().getResolutionResult();
+        try {
+            return resolutionResult.getAllComponents().stream()
+                    .map(ResolvedComponentResult::getId)
+                    .filter(cid -> !cid.equals(resolutionResult.getRoot().getId()))
+                    .filter(ModuleComponentIdentifier.class::isInstance)
+                    .map(ModuleComponentIdentifier.class::cast)
+                    .map(mcid -> ResolvedModule.of(configuration.getName(), mcid.getGroup() + ":" + mcid.getModule()));
```
felixdesouza: this is confusing, module means something in gradle land, but you've taken it and combined it with the group and still called it module.

Either keep it split by adding a new `group` field in `ResolvedModule` or name it something else

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
```diff
@@ -0,0 +1,167 @@
...
+                                            .flatMap(CheckUnusedConstraintsProjectPlugin::resolvedModules)
+                                            .collect(Collectors.toSet())));
+
+                    task.getDependentProjectModule()
+                            .from(incomingDependentProjectModules.map(config -> config.getIncoming()
+                                    .artifactView(view -> view.lenient(true))
+                                    .getFiles()));
+                });
+
+        getConfigurations().register("checkUnusedConstraintsOutgoing", outgoing -> {
```
felixdesouza: let's use the canonical terminology for configurations please: i.e. 
* consumable
* resolvable
* dependencyScope

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
```diff
@@ -0,0 +1,167 @@
...
+        }
+        return projectDependency.getPath();
+    }
+
+    private static Stream<ResolvedModule> resolvedModules(Configuration configuration) {
+        ResolutionResult resolutionResult = configuration.getIncoming().getResolutionResult();
+        try {
+            return resolutionResult.getAllComponents().stream()
+                    .map(ResolvedComponentResult::getId)
+                    .filter(cid -> !cid.equals(resolutionResult.getRoot().getId()))
```
felixdesouza: why do we do this?

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
```diff
@@ -0,0 +1,167 @@
...
+                    incoming.setTransitive(false);
+                    incoming.attributes(attrs -> {
+                        attrs.attribute(
+                                Usage.USAGE_ATTRIBUTE, getObjectFactory().named(Usage.class, OUTGOING_USAGE));
+                    });
+
+                    incoming.withDependencies(deps -> {
+                        getDependentProjectPaths(project)
+                                .forEach(path -> deps.add(getDependencyHandler().project(Map.of("path", path))));
+                    });
```
felixdesouza: why do we need to use `withDependencies` here? is it not enough to declare a dependency on the project path?

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsTask.java
```diff
@@ -150,6 +117,14 @@ public final void checkNoUnusedPin() {
                 "./gradlew checkUnusedConstraints --fix");
     }
 
+    private static Set<ResolvedModule> readModulesFile(File file) {
+        try {
+            return OBJECT_MAPPER.readValue(file, SET_OF_RESOLVED_MODULES);
```
felixdesouza: nit, just inline the type reference, you can just do `new TypeReference<>() {}` doesn't need to be its own constant

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsPlugin.java
```diff
@@ -0,0 +1,103 @@
...
+                    collected.setVisible(false);
+                    collected.attributes(attrs -> {
+                        attrs.attribute(
+                                Usage.USAGE_ATTRIBUTE,
+                                getObjectFactory()
+                                        .named(Usage.class, CheckUnusedConstraintsProjectPlugin.OUTGOING_USAGE));
+                    });
+
+                    // withDependencies does not seem to work on Gradle 7.x.x with configuration-on-demand so use
+                    // addAllLater instead
```
felixdesouza: `addAllLater` or `addAll` should be the default, `withDependencies` should be used as a last resort. So I would not even comment on this, and I would try to ensure that we use this elsewhere

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsPlugin.java
```diff
@@ -0,0 +1,103 @@
...
+                    collected
+                            .getDependencies()
+                            .addAllLater(getProviderFactory().provider(() -> rootProject.getAllprojects().stream()
+                                    .map(subproject ->
+                                            getDependencyHandler().project(Map.of("path", subproject.getPath())))
+                                    .toList()));
+                });
+
+        TaskProvider<CheckUnusedConstraintsTask> checkNoUnusedConstraints = rootProject
```
felixdesouza: let's have the name of the variable match the name of the task please

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
```diff
@@ -0,0 +1,167 @@
...
+    protected abstract ConfigurationContainer getConfigurations();
+
+    @Inject
+    protected abstract DependencyHandler getDependencyHandler();
+
+    @Override
+    public final void apply(Project project) {
+        // Create an incoming configuration that will resolve artifacts from dependent projects.
+        // This establishes proper task ordering through Gradle's dependency resolution system.
+        NamedDomainObjectProvider<Configuration> incomingDependentProjectModules = getConfigurations()
```
felixdesouza: I don't understand why you need this at all?

Each of the resolvable configurations that you get from `GradleConfigurations.getResolvableConfigurations` is resolvable, so if you pass them in correctly, it should be more than enough to get task ordering working correctly.

From the looks of it, we're wrapping it all up in a provider, and thus losing any information that can be gleaned.

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    /**
+     * Returns the set of project paths that this project depends on, excluding self and root projects.
+     *
+     * <p> We must filter these out to avoid cycles where subproject tasks depend on root project tasks.
+     */
+    @SuppressWarnings("deprecation")
+    private static Set<String> getDependentProjectPaths(Project project) {
+        return GradleConfigurations.getResolvableConfigurations(project).stream()
```
felixdesouza: you are effectively doing some form of crude manual dependency resolution here

src/test/java/com/palantir/gradle/versions/CheckUnusedConstraintIntegrationTest.java
```diff
@@ -181,6 +182,157 @@ void unused_check_should_use_exact_matching(GradleInvoker gradle, RootProject ro
...
+
+            configurations.resolvable.dependencyConstraints.addAllLater(
+                provider {
+                    // When this provider is evaluated (during resolvable resolution),
+                    // it resolves the subproject's integrationTestRuntimeClasspath from the root's thread
+                    def integTestClasspath = project(':serviceA')
+                        .configurations
+                        .getByName('integrationTestRuntimeClasspath')
+                    integTestClasspath.resolvedConfiguration.resolvedArtifacts
+                    return []
```
felixdesouza: It's not clear to me how this is representative? this is actually *triggering* a resolution in a constraint as opposed to just referencing something in a different project.

Is there an actual failure that this test accurately encodes?

If there *is* something and it's inside GCV, we should try and fix that rather than include a workaround and then subsequently declare 9.x compatability. If another plugin is doing something like this, then they're not compatible with 9.x of gradle, and we should go and fix them as well.

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
FinlayRJW: split out and renamed to `ResolvedCoordinate`

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsPlugin.java
FinlayRJW: renamed to `checkUnusedConstraintsAllResolvable`

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
FinlayRJW: renamed to `checkUnusedConstraintsAllResolvable` `checkUnusedConstraintsResolvable` `checkUnusedConstraintsConsumable`

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
FinlayRJW: ah this is a hangover from when we were applying all from root - https://github.com/palantir/gradle-consistent-versions/pull/1573/changes#diff-0179967b86cd5c0d944dfbd926cbb844d00ee3ae62abec64eea9ac9812235f4eL186 have removed

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsPlugin.java
FinlayRJW: renamed - looks like it was a[ hangover from copy pasting!](https://github.com/palantir/gradle-consistent-versions/pull/1573/changes#diff-42c32aef39bdeaa13a1fb03d2aa1645f586080dcc77ee7155f27638df1229d7eL76)

src/test/java/com/palantir/gradle/versions/CheckUnusedConstraintIntegrationTest.java
FinlayRJW: so this is idiomatic of what I saw when testing on internal plugins (at least if I understood it correctly) - I assumed the issue was here as the problem only manifest when running `checkUnusedConstraints` 

Honestly my current solution is very brittle I think we either say this is a bug in our other plugins or we force `writeResolvedCoordinatesTask` to run sequentially (which I don't think will be any slower than previous)

FinlayRJW: e.g. I think this is an example internally https://pl.ntr/2zq

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
FinlayRJW: 🤦 I was making this much more complex than needed - so I needed to wrap `GradleConfigurations.getResolvableConfigurations` in a provider else I would miss some configurations but it was the `.flatMap(CheckUnusedConstraintsProjectPlugin::resolvedModules)` that was causing it to lose all task ordering! I've added a second input:

```java
task.getResolvableConfigurationFiles()
                            .from(getProviderFactory()
                                    .provider(() -> GradleConfigurations.getResolvableConfigurations(project).stream()
                                            .filter(config -> !config.getName().startsWith("checkUnusedConstraints"))
                                            .map(config -> config.getIncoming()
                                                    .artifactView(view -> view.lenient(true))
                                                    .getFiles())
                                            .collect(Collectors.toList())));
```

And this seems to be working perfectly! I should have realised this so much sooner! Well at least now I know

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
FinlayRJW: see https://github.com/palantir/gradle-consistent-versions/pull/1573#discussion_r2773462990

src/test/java/com/palantir/gradle/versions/CheckUnusedConstraintIntegrationTest.java
FinlayRJW: see https://github.com/palantir/gradle-consistent-versions/pull/1573#discussion_r2773462990

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
FinlayRJW: could also do a `mustRunAfter` if that is neater

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
```diff
@@ -0,0 +1,98 @@
...
+                    task.getResolvedCoordinates()
+                            .set(GradleConfigurations.getResolvableConfigurations(project)
+                                    .map(configurations -> configurations.stream()
+                                            .flatMap(CheckUnusedConstraintsProjectPlugin::resolvedCoordinateStream)
+                                            .collect(Collectors.toSet())));
+
+                    task.getResolvableConfigurationFiles()
+                            .from(GradleConfigurations.getResolvableConfigurations(project)
+                                    .map(configurations -> configurations.stream()
+                                            .filter(config -> !config.getName().startsWith("checkUnusedConstraints"))
```
FinlayRJW: filter out self to avoid cycles

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
```diff
@@ -0,0 +1,85 @@
...
+                                    .map(configurations -> configurations.stream()
+                                            .filter(configuration ->
+                                                    !configuration.getName().startsWith("checkUnusedConstraints"))
+                                            .map(configuration -> configuration
+                                                    .getIncoming()
+                                                    .artifactView(view -> view.lenient(true))
+                                                    .getFiles())
+                                            .collect(Collectors.toList())));
+
+                    task.getRootComponents()
```
FinlayRJW: I tried using just `getResolutionResult().getRootComponent()` as the input but it seems that gradle needs the files to actually figure out the cross project running order

src/main/java/com/palantir/gradle/versions/WriteResolvedCoordinatesTask.java
```diff
@@ -0,0 +1,73 @@
...
+    @PathSensitive(PathSensitivity.NONE)
+    public abstract ConfigurableFileCollection getResolvedFiles();
+
+    @Input
+    public abstract MapProperty<String, ResolvedComponentResult> getRootComponents();
+
+    @TaskAction
+    public final void writeResolvedCoordinates() {
+        List<ResolvedCoordinate> sorted = getRootComponents().get().entrySet().stream()
+                .flatMap(entry -> DependencyGraphUtils.allComponentResultsFromRoot(entry.getValue()).stream()
```
FinlayRJW, line 56: this was super useful! `DependencyGraphUtils.allComponentResultsFromRoot`

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
```diff
@@ -16,152 +16,70 @@
...
 import org.gradle.api.model.ObjectFactory;
-import org.gradle.api.provider.ProviderFactory;
+import org.gradle.api.tasks.TaskContainer;
 import org.gradle.api.tasks.TaskProvider;
-import org.gradle.util.GradleVersion;
 
 public abstract class CheckUnusedConstraintsProjectPlugin implements Plugin<Project> {
 
-    static final String OUTGOING_USAGE = "checkUnusedConstraintsOutgoingModuleIdentifiers";
+    static final String OUTGOING_USAGE = "gcv-check-unused-constraints";
```
felixdesouza: let's just call this `USAGE`

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsTask.java
```diff
@@ -78,11 +76,11 @@ public CheckUnusedConstraintsTask() {
     @TaskAction
     public final void checkNoUnusedPin() {
         Set<String> excludedConfigs = getExcludeConfigurations().get();
-        Set<String> artifacts = getResolvedModulesFiles().getFiles().stream()
+        Set<String> artifacts = getResolvedCoordinatesFiles().getFiles().stream()
                 .map(CheckUnusedConstraintsTask::readModulesFile)
                 .flatMap(Set::stream)
                 .filter(module -> !excludedConfigs.contains(module.configuration()))
-                .map(ResolvedModule::module)
+                .map(coord -> coord.group() + ":" + coord.module())
```
felixdesouza: full name please: `coordinate`

src/main/java/com/palantir/gradle/versions/WriteResolvedCoordinatesTask.java
```diff
@@ -18,44 +18,54 @@
...
     @TaskAction
-    public final void writeResolvedModules() {
+    public final void writeResolvedCoordinates() {
+        List<ResolvedCoordinate> sorted = getRootComponents().get().entrySet().stream()
+                .flatMap(entry -> DependencyGraphUtils.allComponentResultsFromRoot(entry.getValue()).stream()
+                        .map(ResolvedComponentResult::getId)
+                        .filter(ModuleComponentIdentifier.class::isInstance)
+                        .map(ModuleComponentIdentifier.class::cast)
+                        .map(mcid -> ResolvedCoordinate.of(entry.getKey(), mcid.getGroup(), mcid.getModule())))
+                .sorted(Comparator.comparing(ResolvedCoordinate::configuration)
```
felixdesouza: nit, just make `ResolvedCoordinate` comparable

src/main/java/com/palantir/gradle/versions/WriteResolvedCoordinatesTask.java
felixdesouza: yeah, nice addition

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
```diff
@@ -16,152 +16,70 @@
...
+                                            .filter(configuration ->
+                                                    !configuration.getName().startsWith("checkUnusedConstraints"))
+                                            .map(configuration -> configuration
+                                                    .getIncoming()
+                                                    .artifactView(view -> view.lenient(true))
+                                                    .getFiles())
+                                            .collect(Collectors.toList())));
+
+                    task.getRootComponents()
+                            .putAll(GradleConfigurations.getResolvableConfigurations(project)
```
felixdesouza: can we store this in a varibale and then refer to it? it's kinda difficult to parse with the nesting

felixdesouza: shouldn't you be filtering out the `checkUnusedConstraints` thing as well here?

probably worth storing that to a variable, and then referring to it at both places

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
```diff
@@ -16,152 +16,70 @@
...
+                    task.getResolvedFiles()
+                            .from(GradleConfigurations.getResolvableConfigurations(project)
+                                    .map(configurations -> configurations.stream()
+                                            .filter(configuration ->
+                                                    !configuration.getName().startsWith("checkUnusedConstraints"))
+                                            .map(configuration -> configuration
+                                                    .getIncoming()
+                                                    .artifactView(view -> view.lenient(true))
+                                                    .getFiles())
+                                            .collect(Collectors.toList())));
```
felixdesouza: nit: `collect(Collectors.toList())` -> `toList()`

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
```diff
@@ -16,152 +16,70 @@
...
-                                            .collect(Collectors.toSet())));
-
-                    task.getDependentProjectModule()
-                            .from(incomingDependentProjectModules.map(config -> config.getIncoming()
-                                    .artifactView(view -> view.lenient(true))
-                                    .getFiles()));
+                            .fileValue(new File(task.getTemporaryDir(), "resolved-module-identifiers.json"));
+
+                    task.getResolvedFiles()
+                            .from(GradleConfigurations.getResolvableConfigurations(project)
```
felixdesouza: same thing here, can you put what's in the `from` in its own variable pls

FinlayRJW: split out

src/test/java/com/palantir/gradle/versions/CheckUnusedConstraintIntegrationTest.java
```diff
@@ -181,6 +182,75 @@ void unused_check_should_use_exact_matching(GradleInvoker gradle, RootProject ro
         buildSucceed(gradle);
     }
 
+    @Test
+    @AdditionallyRunWithGradle("9.3.0")
```
felixdesouza: if this is supposed to work with gradle 9, then I feel like we should add gradle 9 to the list of versions rather than trying to add them for specific versions?

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
FinlayRJW: have split out to a var both reference

src/test/java/com/palantir/gradle/versions/CheckUnusedConstraintIntegrationTest.java
FinlayRJW: can do not sure if all tests / parts of GCV are gradle 9 compatible and we still have a split between groovy and java tests

src/main/java/com/palantir/gradle/versions/WriteResolvedCoordinatesTask.java
```diff
@@ -0,0 +1,84 @@
...
+                        return DependencyGraphUtils.allComponentResultsFromRoot(entry.getValue()).stream()
+                                .map(ResolvedComponentResult::getId)
+                                .filter(ModuleComponentIdentifier.class::isInstance)
+                                .map(ModuleComponentIdentifier.class::cast)
+                                .map(mcid -> ResolvedCoordinate.of(entry.getKey(), mcid.getGroup(), mcid.getModule()));
+                    } catch (RuntimeException e) {
+                        log.debug(
+                                "Skipping configuration '{}' due to unresolved dependencies: {}",
+                                entry.getKey(),
+                                e.getMessage());
```
felixdesouza: doesn't this make the result nonsensical?

if you fail to resolve in here, then you could have an actual coordinate that won't get matched, and it will throw errors in the `CheckUnusedConstraintsTask`, in a way that's confusing for the user. Especially considering that it ignores all configurations from this project.

If they then remove said _one_ of the coordinates, and somehow they fix the resolution from a different coordinate, they will get errors trying to resolve the component that was removed.

if we can't resolve, we can't get a meaningful result out of `CheckUnusedConstraints`. You can probably error if you'd like and say that you can't do anything do to an unresolved configuration, but silently moving forward makes the result nonsensical.

src/test/java/com/palantir/gradle/versions/CheckUnusedConstraintIntegrationTest.java
felixdesouza: with the exception of parallel, which I think we should probably add into the testing plugin as a default or using @crogoz new decorator, it should be gradle 9 compatible from my earlier work in the summer last year.

we can keep this for now, but let's FLUP on getting the rest of the tests moved over and flipping on the flags so we don't have to look at this repo for gradle 9 anymore.

FinlayRJW: cool - I've bumped so that whole test class is at least checking gradle 9 as it should all be working

src/main/java/com/palantir/gradle/versions/WriteResolvedCoordinatesTask.java
FinlayRJW: yeah I think you might be right - I think the error message from allComponentResultsFromRoot is pretty good - I added this as I saw a failure in a large internal repo due to a dep not resolving 

`Caused by: java.lang.RuntimeException: Failed to resolve org.scala-lang.modules:scala-java8-compat_2.12                                                                                                                                                             `  

but that may have been a fluke

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
FinlayRJW: huh - for some reason splitting out breaks configuration cache in gradle 8

FinlayRJW: will need to leave as 

```java
                    task.getResolvedFiles().from(configurationsToCheck.map(configurations -> configurations.stream()
                            .map(configuration -> configuration
                                    .getIncoming()
                                    .artifactView(view -> view.lenient(true))
                                    .getFiles())
                            **.collect(Collectors.toList())));**
```

felixdesouza: putting it in a variable breaks configuration cache?

src/main/java/com/palantir/gradle/versions/WriteResolvedCoordinatesTask.java
felixdesouza: yes, ignoring the failure is not going to produce good results

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
FinlayRJW: saw 
```
Could not load the value of field `provider` of `org.gradle.internal.serialize.codecs.core.ProviderBackedFileCollectionSpec` bean found in field `__resolvedFiles__` of task `:writeResolvedCoordinatesTask` of type `com.palantir.gradle.versions.WriteResolvedCoordinatesTask`.
```

but isn't happening when I rerun locally?

FinlayRJW: odd gonna try pushing again and see if it works maybe something else was breaking it?

FinlayRJW: okay seems fine now odd!

src/main/java/com/palantir/gradle/versions/WriteResolvedCoordinatesTask.java
CRogers: Why it isn't just in normal Gradle I don't understand: https://github.com/gradle/gradle/issues/26897

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
FinlayRJW: huh so it seems that maybe it was the `.toList`? that was causing it to fail

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsPlugin.java
```diff
@@ -0,0 +1,103 @@
...
+                    resolvable.setVisible(false);
+                    resolvable.attributes(attributes -> {
+                        attributes.attribute(
+                                Usage.USAGE_ATTRIBUTE,
+                                getObjectFactory().named(Usage.class, CheckUnusedConstraintsProjectPlugin.USAGE));
+                    });
+
+                    resolvable
+                            .getDependencies()
+                            .addAllLater(getProviderFactory().provider(() -> rootProject.getAllprojects().stream()
```
felixdesouza: I think you can unwrap this provider, you can't dynamically add a project after the fact

felixdesouza (approved):
thanks for putting this up!

please flup setting 9.3.0 in the `test-versions.yml` or whatever it's called

PR #1576, opened 2026-02-10
Use `getAllComponents()` instead of manual graph traversal in `checkUnusedConstraints`

FinlayRJW:
## Before this PR

The `WriteResolvedCoordinatesTask` used `DependencyGraphUtils.allComponentResultsFromRoot()` to walk the dependency graph from each configuration's root component. This method throws on any `UnresolvedDependencyResult`, causing `checkUnusedConstraints` to fail in projects that have unresolvable dependencies.

The old inline implementation in `CheckUnusedConstraintsTask.getResolvedModuleIdentifiers` used `ResolutionResult.getAllComponents()` which is lenient — it only returns successfully resolved components and silently omits unresolved ones see:

```java
    public static void eachElement(
        ResolvedComponentResult start,
        Action<? super ResolvedComponentResult> moduleAction,
        Action<? super DependencyResult> dependencyAction,
        Set<ResolvedComponentResult> visited
    ) {
        if (!visited.add(start)) {
            return;
        }
        moduleAction.execute(start);
        for (DependencyResult d : start.getDependencies()) {
            dependencyAction.execute(d);
            if (d instanceof ResolvedDependencyResult) {
                eachElement(((ResolvedDependencyResult) d).getSelected(), moduleAction, dependencyAction, visited);
            }
        }
    }
```

## After this PR

Uses `ResolutionResult.getAllComponents()` instead of manual graph traversal, restoring the lenient behaviour. This also removes the `dependency-graph-utils` dependency which is no longer needed.

This is fine todo as we are wrapping in a provider anyway so the resolution is still lazy (we were using `getRoot` previously which is not lazy anyway)

## Testing

Existing tests cover `checkUnusedConstraints` behavior.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -64,7 +64,6 @@ dependencies {
     implementation 'com.fasterxml.jackson.datatype:jackson-datatype-guava'
     implementation 'one.util:streamex'
     implementation 'com.palantir.gradle.idea-configuration:gradle-idea-configuration'
-    implementation 'com.palantir.gradle.utils:dependency-graph-utils'
 
     testImplementation platform('org.junit:junit-bom')
     testImplementation 'com.netflix.nebula:nebula-test'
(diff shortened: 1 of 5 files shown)


felixdesouza:
👍

PR #1578, opened 2026-02-10
Convert `GetVersionPluginSpec` to gradle-plugin-testing framework

FinlayRJW:
## Before this PR

`GetVersionPluginSpec` was a Groovy Spock test extending Nebula's `ProjectSpec`, which is being replaced by the `gradle-plugin-testing` framework.

## After this PR

Converts `GetVersionPluginSpec` (Groovy/Spock/Nebula `ProjectSpec`) → `GetVersionPluginIntegrationTest` (Java/JUnit 5/`@GradlePluginTests`).

The 3 `getVersion` callable tests are preserved:
- `getVersion` callable with `(String, Configuration)` args
- `getVersion` callable with `(String, String, Configuration)` args
- `getVersion` callable with `(String, String)` args (uses `unifiedClasspath`)

The two `apply` tests (`apply does not throw exceptions`, `apply is idempotent`) are dropped as they were trivial and don't add value as integration tests.

Since `GetVersionPlugin` doesn't have its own plugin ID (it's only applied internally by `ConsistentVersionsPlugin`), the tests apply `com.palantir.consistent-versions` as the workaround for no `ProjectSpec` equivalent. The `getVersion` Groovy closure is exercised via custom tasks that invoke `project.ext['getVersion'](...)` at execution time.

## Possible alternative approaches?

Could register a dedicated plugin ID for `GetVersionPlugin` to allow direct testing, but that would change the public API surface unnecessarily.


src/test/java/com/palantir/gradle/versions/GetVersionPluginIntegrationTest.java
```diff
@@ -0,0 +1,89 @@
...
+import com.palantir.gradle.testing.execution.GradleInvoker;
+import com.palantir.gradle.testing.execution.InvocationResult;
+import com.palantir.gradle.testing.junit.DisabledConfigurationCache;
+import com.palantir.gradle.testing.junit.GradlePluginTests;
+import com.palantir.gradle.testing.project.RootProject;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+@GradlePluginTests
+@DisabledConfigurationCache
```
felixdesouza: please add a why?

src/test/java/com/palantir/gradle/versions/GetVersionPluginIntegrationTest.java
```diff
@@ -0,0 +1,89 @@
...
+    @Test
+    void get_version_is_callable_from_groovy_with_string_and_configuration_args(
+            GradleInvoker gradle, RootProject project) {
+        project.buildGradle().plugins().add("java");
+        project.buildGradle().append("""
+            tasks.register('callGetVersion') {
+                doLast {
+                    project.ext['getVersion']('com.google.guava:guava', configurations.runtimeClasspath)
+                }
+            }
```
crogoz, line 49: I think you can call directly getVersion. Same below

```suggestion
            tasks.register('callGetVersion') {
                getVersion('com.google.guava:guava', configurations.runtimeClasspath)
            }
```

FinlayRJW: thanks! didn't know you could do that - fixed in [5d16a45](https://github.com/palantir/gradle-consistent-versions/pull/1578/commits/5d16a45a46d40c34037d040ff36b170316334bb7)

crogoz (approved):

PR #1580, opened 2026-02-12
Filter artifact view to `ModuleComponentIdentifier` in `checkUnusedConstraints`

FinlayRJW:
## Before this PR

In multi-project builds with inter-project dependencies, `checkUnusedConstraints` fails because `writeResolvedCoordinatesTask` uses the output of subproject `:jar` tasks (the resolved artifact JARs) without declaring an explicit dependency on them. Gradle detects this implicit dependency and fails the build with:

```
Task ':writeResolvedCoordinatesTask' uses this output of task ':some-subproject:jar'
without declaring an explicit or implicit dependency.
```

This was introduced in #1573 which changed the resolved files input to use an artifact view. The artifact view resolves files for _all_ components including project dependencies, which produces JAR files from other subprojects. 

## After this PR

Adds a `componentFilter` to the artifact view that filters to only `ModuleComponentIdentifier` components, matching the downstream filtering in `WriteResolvedCoordinatesTask`. This is safe because:

1. The `componentFilter` operates at the artifact selection level, not graph resolution — Gradle must fully resolve the configuration's dependency graph before applying the filter, so the task ordering that prevents parallel resolution lock errors is preserved
2. `WriteResolvedCoordinatesTask` already filters to only `ModuleComponentIdentifier` when extracting coordinates — project components would be skipped anyway
3. Project dependencies are not version-constrained via `versions.props`, so they're irrelevant to the unused constraints check

The filter removes the implicit build task dependency on project `:jar` tasks (fixing the error) while preserving the configuration resolution ordering needed for correct parallel execution.

## Possible downsides?

### DIFF
--- a/src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
@@ -26,6 +26,7 @@ import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.artifacts.Configuration;
 import org.gradle.api.artifacts.ConfigurationContainer;
+import org.gradle.api.artifacts.component.ProjectComponentIdentifier;
 import org.gradle.api.artifacts.result.ResolvedComponentResult;
 import org.gradle.api.attributes.Usage;
 import org.gradle.api.file.FileCollection;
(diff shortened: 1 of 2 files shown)


src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
```diff
@@ -70,7 +71,10 @@ public final void apply(Project project) {
                 configurationsToCheck.map(configurations -> configurations.stream()
                         .map(configuration -> configuration
                                 .getIncoming()
-                                .artifactView(view -> view.lenient(true))
+                                .artifactView(view -> {
+                                    view.lenient(true);
+                                    view.componentFilter(ModuleComponentIdentifier.class::isInstance);
```
CRogers, line 76: Can we write a failing test for this case?

FinlayRJW: I've added a test that demonstrates the issue - could not reproduce the exact failure 
```
Task ':writeResolvedCoordinatesTask' uses this output of task ':some-subproject:jar'
without declaring an explicit or implicit dependency.
```
Not sure if conjure does something weird with dialogue subprojects?

CRogers (approved):

PR #1588, opened 2026-02-16
Stop downloading artifacts in `checkUnusedConstraints` resolution ordering

FinlayRJW:
## Before this PR

The `writeResolvedCoordinatesTask` in #1573 used `@InputFiles` with an artifact view filtered to `ModuleComponentIdentifier` to establish cross-project resolution ordering for `--parallel` on Gradle 9. This works correctly for locking, but the artifact view with `componentFilter(ModuleComponentIdentifier.class::isInstance)` selects all external module artifacts, causing Gradle to **download every JAR** from every resolvable configuration (compileClasspath, runtimeClasspath, devEnv, custom configurations, etc.).

This caused significant performance regression:
- Tasks that previously took 7-20 seconds (metadata-only resolution) now take 40-50+ seconds
- The actual task logic only needs the dependency graph metadata (module coordinates), not the artifact files

## After this PR

Changes the `componentFilter` from `ModuleComponentIdentifier.class::isInstance` (selects all external module artifacts → downloads JARs) to `_id -> false` (selects nothing → downloads no artifacts).

The `@InputFiles` file collection is still evaluated by Gradle during input fingerprinting, which triggers dependency graph resolution with proper cross-project locking. The component filter is applied *after* the graph is resolved, so even though no components are selected (and no files are downloaded), the resolution still happens — establishing the ordering needed for `--parallel` on Gradle 9.

The `@Input MapProperty<String, Set<ResolvedComponentResult>>` then provides the actual dependency graph data. Since the graph was already resolved via the `@InputFiles` evaluation, the `getAllComponents()` calls return cached results.

## Possible downsides?

The `@InputFiles` file collection will always be empty (since `componentFilter(_id -> false)` matches nothing). This means Gradle's up-to-date checking for this input will always see "no files" — but the `@Input` with `ResolvedComponentResult` objects still provides proper change detection for the dependency graph.

### DIFF
--- a/src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
@@ -72,6 +72,10 @@ public abstract class CheckUnusedConstraintsProjectPlugin implements Plugin<Proj
                                 .getIncoming()
                                 .artifactView(view -> {
                                     view.lenient(true);
+                                    // Exclude all components so no artifact files are downloaded.
+                                    // The artifact view still triggers dependency graph resolution
+                                    // (the filter is applied after resolution), which is all we need
+                                    // for cross-project locking with --parallel on Gradle 9+.
                                     view.componentFilter(_id -> false);
                                 })
                                 .getFiles())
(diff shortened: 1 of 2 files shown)


FinlayRJW:
Tested an speed seems to be back to previous speeds and not seeing the locking issue in gradle 9

src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
```diff
@@ -73,7 +72,11 @@ public final void apply(Project project) {
                                 .getIncoming()
                                 .artifactView(view -> {
                                     view.lenient(true);
-                                    view.componentFilter(ModuleComponentIdentifier.class::isInstance);
+                                    // Exclude all components so no artifact files are downloaded.
+                                    // The artifact view still triggers dependency graph resolution
+                                    // (the filter is applied after resolution), which is all we need
+                                    // for cross-project locking with --parallel on Gradle 9+.
+                                    view.componentFilter(_id -> false);
```
FinlayRJW, line 79: this feels like a bit of a work-around not sure if there is a more gradle-y way of preventing the cross-project locking

PR #1589, opened 2026-02-16
Finlayw/try single conf

FinlayRJW:
## After this PR

==COMMIT_MSG==
==COMMIT_MSG==

### DIFF
--- a/src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
@@ -72,6 +72,10 @@ public abstract class CheckUnusedConstraintsProjectPlugin implements Plugin<Proj
                                 .getIncoming()
                                 .artifactView(view -> {
                                     view.lenient(true);
+                                    // Exclude all components so no artifact files are downloaded.
+                                    // The artifact view still triggers dependency graph resolution
+                                    // (the filter is applied after resolution), which is all we need
+                                    // for cross-project locking with --parallel on Gradle 9+.
                                     view.componentFilter(_id -> false);
                                 })
                                 .getFiles())
--- a/src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
@@ -22,13 +22,13 @@ import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 import javax.inject.Inject;
+import org.gradle.api.NamedDomainObjectProvider;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.artifacts.Configuration;
 import org.gradle.api.artifacts.ConfigurationContainer;
 import org.gradle.api.artifacts.result.ResolvedComponentResult;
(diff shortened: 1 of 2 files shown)


src/main/java/com/palantir/gradle/versions/CheckUnusedConstraintsProjectPlugin.java
```diff
@@ -67,22 +77,13 @@ public final void apply(Project project) {
...
-                                })
-                                .getFiles())
-                        .collect(Collectors.toList()));
-
         TaskProvider<WriteResolvedCoordinatesTask> writeResolvedCoordinatesTask = getTasks()
                 .register("writeResolvedCoordinatesTask", WriteResolvedCoordinatesTask.class, task -> {
                     task.getOutputFile()
                             .fileValue(new File(task.getTemporaryDir(), "resolved-module-identifiers.json"));
-                    task.getResolvedFiles().from(resolvedFiles);
+                    task.getOrderingResult()
```
FinlayRJW, line 84: this has the same locking error - seemingly `getRootComponent` is not passing through the order information gradle needs?

Using:

```java
task.getOrderingResult().from(orderingConfiguration.map(conf -> conf.getIncoming()
                            .artifactView(view -> {
                                view.lenient(true);
                                view.componentFilter(_id -> false);
                            })
                            **.getFiles()));**
```
works but that is not really any different from #1588

FinlayRJW: The [javadoc for `getRootComponent()`](https://github.com/gradle/gradle/blob/c77c173180bfd16f6ea361089f09841a94ddf656/subprojects/core-api/src/main/java/org/gradle/api/artifacts/result/ResolutionResult.java#L49) claims it "tracks the producer tasks", but as we can see it returns a plain [`DefaultProvider`](https://github.com/gradle/gradle/blob/c77c173180bfd16f6ea361089f09841a94ddf656/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/resolver/DefaultResolutionOutputs.java#L91-L94) which inherits [`ValueProducer.unknown()`](https://github.com/gradle/gradle/blob/c77c173180bfd16f6ea361089f09841a94ddf656/platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/provider/AbstractMinimalProvider.java#L135-L137) — a [no-op](https://github.com/gradle/gradle/blob/c77c173180bfd16f6ea361089f09841a94ddf656/platforms/core-configuration/model-core/src/main/java/org/gradle/api/internal/provider/ValueSupplier.java#L160-L169) that carries no task dependency information.

Unlike `getRootComponent()`, `FileCollection` from artifact views goes through [`ResolutionBackedFileCollection.visitDependencies()`](https://github.com/gradle/gradle/blob/c77c173180bfd16f6ea361089f09841a94ddf656/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/configurations/ResolutionBackedFileCollection.java#L56-L58) → [`getTaskDependencyValue()`](https://github.com/gradle/gradle/blob/c77c173180bfd16f6ea361089f09841a94ddf656/platforms/software/dependency-management/src/main/java/org/gradle/api/internal/artifacts/configurations/ResolutionResultProvider.java#L32-L36), which resolves project dependencies and adds their artifact-producing tasks to the task graph — establishing cross-project ordering.

PR #1595, opened 2026-02-19
Fix OOM in `writeResolvedCoordinatesTask` by avoiding serialisation of full resolution graph

FinlayRJW:
## Before this PR

In large repos with many subprojects, `writeResolvedCoordinatesTask` OOMs the Gradle daemon. The task declares `MapProperty<String, Set<ResolvedComponentResult>>` as an `@Input`. When Gradle fingerprints this input for up-to-date checking, it recursively serialises the entire dependency resolution graph per configuration per subproject — not just the component IDs the task actually needs.

JFR profiling confirmed this: the dominant allocation path is `AbstractValueProcessor.gradleSerialized` → `ResolvedComponentResultSerializer.writeComponent` (recursive), with individual worker threads allocating 600MB–2.4GB each before the daemon is killed.

## After this PR

==COMMIT_MSG==
Fix `writeResolvedCoordinatesTask` OOM by extracting lightweight `ResolvedCoordinate` values in the provider instead of passing `ResolvedComponentResult` objects as task `@Input`. Gradle now fingerprints small string tuples instead of recursively serialising entire resolution graphs.
==COMMIT_MSG==

Tested on the failing repo and now passing, all UT still pass 

## Possible downsides?

### DIFF
--- a/src/main/java/com/palantir/gradle/versions/ResolvedCoordinate.java
+++ b/src/main/java/com/palantir/gradle/versions/ResolvedCoordinate.java
@@ -25,7 +25,7 @@ import org.immutables.value.Value;
 @Value.Immutable
 @JsonSerialize(as = ImmutableResolvedCoordinate.class)
 @JsonDeserialize(as = ImmutableResolvedCoordinate.class)
-public interface ResolvedCoordinate extends Serializable, Comparable<ResolvedCoordinate> {
+interface ResolvedCoordinate extends Serializable, Comparable<ResolvedCoordinate> {
     Comparator<ResolvedCoordinate> COMPARATOR = Comparator.comparing(ResolvedCoordinate::configuration)
             .thenComparing(ResolvedCoordinate::module)
             .thenComparing(ResolvedCoordinate::group);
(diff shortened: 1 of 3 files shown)


src/main/java/com/palantir/gradle/versions/ResolvedCoordinate.java
```diff
@@ -25,7 +25,7 @@
 @Value.Immutable
 @JsonSerialize(as = ImmutableResolvedCoordinate.class)
 @JsonDeserialize(as = ImmutableResolvedCoordinate.class)
-interface ResolvedCoordinate extends Serializable, Comparable<ResolvedCoordinate> {
+public interface ResolvedCoordinate extends Serializable, Comparable<ResolvedCoordinate> {
```
felixdesouza, line 28: why is this public now?

FinlayRJW: ResolvedCoordinate is now an input to a task so is public right? without it I see `ResolvedCoordinate` is exposed beyond its scope in intellij

felixdesouza: I think gradle is fine with it, just that intellij is not, and that's correct. I would have probably just made all the tasks non-public, but that ship has sailed.

felixdesouza:
👍

PR #1614, opened 2026-03-10
Skip `extendsFrom(rootConfiguration)` for consumable-only configurations

FinlayRJW:
## Before this PR

GCV applies `conf.extendsFrom(rootConfiguration)` to **all** configurations on every project. `rootConfiguration` contains a `ProjectDependency` on the root project. This creates a variant model cycle when two conditions are met on the root project:

1. A **consumable-only** configuration exists (e.g. created by a plugin like `gradle-idea-language-injector`)
2. That consumable config's outgoing artifacts are backed by a **provider** that resolves another configuration (e.g. `compileClasspath`)

Both configurations extend `rootConfiguration` (via GCV). On Gradle 9.4.0, [gradle/gradle#36245](https://github.com/gradle/gradle/pull/36245) changed `extendsFrom` from eagerly storing parent configs in a `Set<Configuration>` to lazily storing them as `Provider<Configuration>`. When Gradle computes the root project's variant model, it now iterates consumable configs and evaluates their lazy `extendsFrom` chains. Following `rootConfiguration` → `ProjectDependency(root)` → re-enters variant model computation on root → cycle.

The real-world trigger is [`gradle-idea-language-injector`](https://github.com/palantir/gradle-idea-language-injector), which creates a consumable config on the root project whose outgoing artifacts resolve `compileClasspath` to scan jars for annotations. See [palantir/gradle-idea-language-injector#149](https://github.com/palantir/gradle-idea-language-injector/pull/149) for a downstream workaround.

## After this PR

Skips `extendsFrom(rootConfiguration)` for consumable-only configurations (`canBeConsumed=true`, `canBeResolved=false`). These are published variants that never resolve dependencies, so inheriting version constraints serves no purpose.

No behavior change for existing configs:
- `implementation`, `api`, etc. — `canBeConsumed=false` → still gets `extendsFrom`
- `compileClasspath`, `runtimeClasspath` — `canBeResolved=true` → still gets `extendsFrom`
- `apiElements`, `runtimeElements` — already handled by `JAVA_PUBLISHED_CONFIGURATION_NAMES`

## Possible downsides?
We are no longer extending from rootConfiguration for any consumable configuration - this could have some unseen side effect, but I don't think it should as we don't care about version constraints from the rootConfiguration on a consumable as they are never resolved
## Testing

### DIFF
--- a/src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
@@ -145,6 +145,7 @@ public abstract class VersionsPropsPlugin implements Plugin<Project> {
         project.subprojects(subproject -> subproject.getPluginManager().apply(VersionsPropsPlugin.class));
     }
 
+    @SuppressWarnings("checkstyle:CyclomaticComplexity")
     private static void setupConfiguration(
             Project subproject,
             VersionRecommendationsExtension extension,
--- a/src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
@@ -176,7 +176,7 @@ public abstract class VersionsPropsPlugin implements Plugin<Project> {
         // Consumable-only configurations are published variants of a project — they don't resolve
         // dependencies, so version constraints from rootConfiguration are useless on them. Extending
         // rootConfiguration also creates a variant model cycle when rootConfiguration contains a
-        // ProjectDependency on the root project (see https://github.com/gradle/gradle/pull/36245).
+        // ProjectDependency on the root project
         if (conf.isCanBeConsumed() && !conf.isCanBeResolved()) {
             log.debug("Not configuring consumable-only configuration: {}", conf);
             return;
--- a/src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
@@ -174,11 +174,14 @@ public abstract class VersionsPropsPlugin implements Plugin<Project> {
             return;
         }
 
-        // Consumable-only configurations are published variants of a project — they never resolve
-        // dependencies, so version constraints from rootConfiguration are useless on them.
-        // Additionally, extending rootConfiguration here creates a variant model cycle on
-        // Gradle 9.4.0+. If a consumable config's outgoing artifacts provider resolves another config (e.g.
-        // compileClasspath) that also extends rootConfiguration.
+        // Consumable-only configurations are outgoing variants — they expose artifacts to consumers
+        // but never resolve dependencies themselves, so version constraints inherited from
+        // rootConfiguration have no effect on them.
+        // Additionally, extending rootConfiguration on a consumable-only config creates a variant
+        // model cycle on Gradle 9.4.0+ when that configuration's outgoing artifacts resolve a configuration (e.g.
+        // compileClasspath) that also extends rootConfiguration. rootConfiguration ends up reachable from both sides —
+        // Gradle includes it in the variant model through the consumable config, and also needs to resolve it through
+        // compileClasspath.
         if (conf.isCanBeConsumed() && !conf.isCanBeResolved()) {
             log.debug("Not configuring consumable-only configuration: {}", conf);
             return;
--- a/src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
@@ -173,10 +173,11 @@ public abstract class VersionsPropsPlugin implements Plugin<Project> {
             return;
         }
 
-        // Consumable-only configurations are published variants of a project — they don't resolve
-        // dependencies, so version constraints from rootConfiguration are useless on them. Extending
-        // rootConfiguration also creates a variant model cycle when rootConfiguration contains a
-        // ProjectDependency on the root project
+        // Consumable-only configurations are published variants of a project — they never resolve
+        // dependencies, so version constraints from rootConfiguration are useless on them.
+        // Additionally, extending rootConfiguration here creates a variant model cycle on
+        // Gradle 9.4.0+. If a consumable config's outgoing artifacts provider resolves another config (e.g.
+        // compileClasspath) that also extends rootConfiguration.
         if (conf.isCanBeConsumed() && !conf.isCanBeResolved()) {
             log.debug("Not configuring consumable-only configuration: {}", conf);
             return;
--- a/src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
@@ -173,6 +173,15 @@ public abstract class VersionsPropsPlugin implements Plugin<Project> {
             return;
         }
 
+        // Consumable-only configurations are published variants of a project — they don't resolve
+        // dependencies, so version constraints from rootConfiguration are useless on them. Extending
+        // rootConfiguration also creates a variant model cycle when rootConfiguration contains a
+        // ProjectDependency on the root project (see https://github.com/gradle/gradle/pull/36245).
+        if (conf.isCanBeConsumed() && !conf.isCanBeResolved()) {
+            log.debug("Not configuring consumable-only configuration: {}", conf);
+            return;
+        }
+
         // Must do all this in a withDependencies block so that it's run lazily, so that
         // `extension.shouldExcludeConfiguration` isn't queried too early (before the user had the change to configure).
         // However, we must not make this lazy using an afterEvaluate.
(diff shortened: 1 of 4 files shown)


src/test/java/com/palantir/gradle/versions/VersionsPropsPluginIntegrationTest.java
```diff
@@ -83,7 +84,7 @@ void setup(MavenRepo repo, RootProject rootProject) {
 
             allprojects {
                 repositories {
-                    maven { url "file:///%s" }
+                    maven { url uri("%s") }
```
FinlayRJW: for some reason fails on gradle 9.4.0 without this? 

```
Caused by: java.net.URISyntaxException: Illegal character in path at index 121: file:////Volumes/git/gradle-consistent-versions/build/gradle-plugin-testing/VersionsPropsPluginIntegrationTest/consumable only configuration on root project does not cause variant model cycle/9.4.0/build/mavenrepo/localMavenRepository
	at org.gradle.api.internal.file.FileNotationConverter.convert(FileNotationConverter.java:96)
```

src/test/java/com/palantir/gradle/versions/VersionsPropsPluginIntegrationTest.java
```diff
@@ -298,6 +299,30 @@ void creates_rootconfiguration_even_if_versions_props_file_missing(GradleInvoker
         gradle.withArgs().buildsSuccessfully();
     }
 
+    @Test
+    @AdditionallyRunWithGradle("9.4.0")
```
FinlayRJW, line 305: when all the tests are moved over we can start testing everything on 9.4.0

src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
```diff
@@ -184,46 +194,58 @@ private static void setupConfiguration(
...
-            if (extension.shouldExcludeConfiguration(conf.getName())) {
-                log.debug("Not configuring {} because it's excluded", conf);
-                return;
-            }
+            setupConfigurationDependencies(
+                    subproject, extension, rootConfiguration, versionsProps, conf, deps, wasConfigured);
+        });
+    }
+
+    private static void setupConfigurationDependencies(
```
felixdesouza: this is not a worthwhile change, please undo

src/test/java/com/palantir/gradle/versions/VersionsPropsPluginIntegrationTest.java
felixdesouza: what is said reason?

src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
```diff
@@ -173,6 +173,16 @@ private static void setupConfiguration(
             return;
         }
 
+        // Consumable-only configurations are published variants of a project — they never resolve
+        // dependencies, so version constraints from rootConfiguration are useless on them.
+        // Additionally, extending rootConfiguration here creates a variant model cycle on
+        // Gradle 9.4.0+. If a consumable config's outgoing artifacts provider resolves another config (e.g.
+        // compileClasspath) that also extends rootConfiguration.
```
felixdesouza: * when you say "published" what do you mean?
* > so version constraints from rootConfiguration are useless on them.
  
  I'm not sure I buy this, can you explain your reasoning a bit more please? ~What happens to configurations that are all three types? this happens in a *lot* of places.~ Will they be broken when they get to 9.x? Probably, but I'm concerned about existing behaviour changing in subtle ways for 8.x. This is also for all consumable-only configurations, not just the root project as that was your initial problem.

src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
FinlayRJW: have undone - but now require `@SuppressWarnings("checkstyle:CyclomaticComplexity")` I assume thats okay?

src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
felixdesouza: > If a consumable config's outgoing artifacts provider resolves another config (e.g. compileClasspath) that also extends rootConfiguration.

I can't parse this sentence, what are you trying to say here

src/test/java/com/palantir/gradle/versions/VersionsPropsPluginIntegrationTest.java
FinlayRJW: well gradle 8.14.3 -> gradle 9.4.0 gives:

```
file:////Volumes/git/gradle-consistent-versions/build/gradle-plugin-testing/VersionsPropsPluginIntegrationTest/consumable only configuration on root project does not cause variant model cycle/8.14.3/build/mavenrepo/localMavenRepository
```
vs 
```
file:////Volumes/git/gradle-consistent-versions/build/gradle-plugin-testing/VersionsPropsPluginIntegrationTest/consumable only configuration on root project does not cause variant model cycle/9.4.0/build/mavenrepo/localMavenRepository
```

Whats weird is that the illegal character the error references to is a space which is present in both

FinlayRJW: maybe gradle 9 is stricter on having spaces in url strings?

FinlayRJW: ah yeah see: https://docs.gradle.org/9.4.0/userguide/upgrading_version_8.html#deprecated_invalid_url_decoding

felixdesouza: ~idk, you tell me! can you go and verify that please?~ thank you! what do you think that means for some of our excavators?

src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
FinlayRJW: > when you say "published" what do you mean?

sorry, I mean consumable-only configurations are outgoing variants — they expose artifacts to consumers but never resolve dependencies themselves. Since they never resolve, version constraints inherited from rootConfiguration surely will have no effect on them

> I can't parse this sentence, what are you trying to say here

Sorry I'm not 100% sure I fully understand but my best guess is extending `rootConfiguration` on a consumable-only config creates a variant model cycle on Gradle 9.4.0+ when that configurations outgoing artifacts resolve a configuration (e.g. `compileClasspath`) that also extends `rootConfiguration`. `rootConfiguration` ends up reachable from both sides — Gradle includes it in the variant model through the consumable config, and also needs to resolve it through `compileClasspath`. That circular dependency between variant discovery and resolution is what Gradle 9.4.0+ now rejects.

FinlayRJW: > This is also for all consumable-only configurations

True I'll add a test to see if this can happen only within a subproject or if this is only an issue with root and tighten the check accordingly

felixdesouza: I guess my qualm with this approach is that, underlying behaviour has changed in gradle 9.4, this change happens to avoid this new behaviour you don't want. However, what I don't get from this PR or its description is what the underlying problem is for this change to be deemed the correct fix. It just feels like because it solves the problem that it's then good to go, which I fundamentally disagree with.

FWIW, this might actually be the right fix, but it's hard to know that given what I'm seeing here, especially considering it may change behaviour for non gradle 9 use cases.

src/test/java/com/palantir/gradle/versions/VersionsPropsPluginIntegrationTest.java
FinlayRJW: should be okay we have an explicit callout to use `uri` I added a bit ago:

``` java
// ALSO CORRECT - allprojects block when subprojects also need the repo
rootProject.buildGradle().append("""
    allprojects {
        repositories {
            maven { url uri("%s") }
        }
    }
    """, repo.path());
```
https://github.com/palantir/gradle-plugin-testing/blob/develop/docs/junit-migration-excavator-instructions.md

src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java
FinlayRJW: yeah thats fair - I'll give it some more thought I think I need to more fully understand how this has changed - will close this PR for now / convert to draft and try and figure out exactly what has changed and what the best fix is

src/test/java/com/palantir/gradle/versions/VersionsPropsPluginIntegrationTest.java
felixdesouza: I would not say that it's explicit, it's just an example of adding repositories. If we're seeing this come up again, we should have an explicit callout along the line of:

* don't use `file:///...` when in `build.gradle` test files, use `uri("%s")` instead as it is deprecated

felixdesouza: one could argue, it's a separate concern as it's gradle 9 stuff, but I think it's far easier to nip it in the bud now whilst we're doing the migrations and it migrates to the proper thing, that's forward compatible, as we probably won't run a similar style AI assisted migration for gradle 9 stuffs.

FinlayRJW: sweet - see https://github.com/palantir/gradle-plugin-testing/pull/416

PR #1664, opened 2026-05-27
Fix `getVersion` failing on Gradle 9 with "Resolution of ':unifiedClasspath' was attempted without an exclusive lock"

FinlayRJW:
## Before this PR

`getVersion('group:name')` (added by `GetVersionPlugin`) lets build scripts look up the version GCV has settled on for a dependency, e.g.:

```gradle
task printGuavaVersion {
    doLast { println getVersion('com.google.guava:guava') }
}
```

Under the hood, the no-configuration overloads resolved the **root** project's `unifiedClasspath` configuration and read the resolved version out of the graph.

On **Gradle 9** this breaks. Resolving a configuration that belongs to another project (here, a subproject reaching into the **root** project's `unifiedClasspath`) without holding that project's lock errors

```
Caused by: org.gradle.api.internal.artifacts.configurations.DefaultConfiguration$IllegalResolutionException:
Resolution of the configuration ':unifiedClasspath' was attempted without an exclusive lock. This is unsafe and not allowed.
        at com.palantir.gradle.versions.GetVersionPlugin.getOptionalVersion(GetVersionPlugin.java:98)
        at com.palantir.gradle.versions.GetVersionPlugin.getVersion(GetVersionPlugin.java:83)
        at com.palantir.gradle.versions.GetVersionPlugin$1.doCall(GetVersionPlugin.java:66)
```

This surfaces most commonly because build scripts call `getVersion` themselves, typically to set a product dependency's `minimumVersion` inside `sls-packaging`'s `distribution` extension:

```gradle
// in a subproject's build.gradle
distribution {
    productDependency {
        productGroup = 'com.palantir.foo'
        productName = 'foo-service'
        minimumVersion = getVersion('com.palantir.foo:foo-api')
        maximumVersion = '1.x.x'
    }
}
```
Because that runs in a subproject but resolves the root's `unifiedClasspath`, the corresponding tasks fail on Gradle 9:

```
Execution failed for task ':my-service:resolveProductDependencies'.
> Error while evaluating property 'productDependencies' of task ':my-service:resolveProductDependencies'.
   > ...
      > Error validating product dependency declared from project: {projectName=my-service}
```

But it affects **any** `getVersion('group:name')` call from a subproject, not just `sls-packaging`.

## After this PR

==COMMIT_MSG==
Fixed `getVersion('group:name')` failing on Gradle 9 with `Resolution of the configuration ':unifiedClasspath' was attempted without an exclusive lock` when called from a subproject. `getVersion` now reads the strict version constraints already declared on the root project's `gcvLocks` platform configuration (which GCV populates from `versions.lock`).
==COMMIT_MSG==

The key insight: the *"without an exclusive lock"* error only applies to **resolution**. Reading a configuration's declared `DependencyConstraint`s is allowed.

GCV already publishes every locked version as a strict `DependencyConstraint` on the root project's `gcvLocks` platform configuration (populated from `versions.lock` in `VersionsLockPlugin`). So the no-configuration `getVersion` overloads now **read** the strict version from those constraints rather than **resolving** `unifiedClasspath`

**Why this is safe / equivalent.** `versions.lock` is by definition the serialised resolution of `unifiedClasspath`, and `verifyLocks` (wired into `check`) enforces that the current resolution still matches the lock. The two can only differ when the lock is stale (already a `verifyLocks` failure)

The not-found error message now points at `versions.lock` and mentions a stale lock as a possible cause.

### Alternatives considered

Both `sls-packaging`-side and GCV-side approaches were prototyped; this GCV-side fix was chosen because it's general (covers every `getVersion` caller), requires **no** build-script or DSL migration, and keeps existing scripts working unchanged.

- **`sls-packaging`: `minimumVersionFrom` DSL** (palantir/sls-packaging#2078) - add a `minimumVersionFrom` to the `distribution` extension and lock it via GCV. Works, but building a `Provider<List<Dependency>>` from the declared product dependencies and adding it to a configuration causes the *old* `getVersion` to resolve `unifiedClasspath` at **configuration time**.
- **`sls-packaging`: break the provider chain with `.flatMap`** - a ~2-line change that just makes evaluation less lazy, moving `getVersion` from execution time to (single-threaded, legal) graph-construction time. This felt more like a work-around than a fix
- **`sls-packaging`: resolve once in a root task** - a root task *can* legally resolve `unifiedClasspath` (like `writeLocks`/`verifyLocks` do), write a `resolved-versions.json`, and have subprojects read it as an `@InputFile`. But it breaks the `getVersion('group:name', configuration)` overload, which must resolve the *caller's* configuration, not the root's.
- **GCV: read `versions.lock` from disk directly** - functionally identical to reading `gcvLocks`..
- **GCV: a resolvable locked configuration per `getVersion` call** - doesn't work; it pulls its constraints from the root `gcvLocks` platform, so resolving it from a subproject hits the same exclusive-lock error.

## Possible downsides?

- **Not full Isolated Projects compatibility.** This removes the *resolution-without-a-lock* error, but `getVersion` still reads the root project's configuration model from a subproject (`project.getRootProject().getConfigurations()`), which a fully isolated-projects build would still disallow.
- **`getVersion` is now only as accurate as `versions.lock`.** Previously it reflected a live resolution. In practice the lock is enforced by `verifyLocks`, and where the two could differ (a stale lock) the locked value is what's actually on the classpath.
- **Compatibility edge case:** ~~I couldn't find an in-repo usage~~ found such a case https://pl.ntr/2Bi, but a consumer build that previously called `getVersion('group:name')` for a sibling project coordinate would now fail because the default lookup only reads `versions.lock`, which excludes project components. -> fixed in c940b66fcbfbfa3f011d41daf1994a57b609dfb0

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,9 +208,9 @@ task printGuavaVersion {
 }
 ```
 
-This function may not be invoked at [Gradle Configuration time](https://docs.gradle.org/current/userguide/build_lifecycle.html) as it involves resolving dependencies.  Put it inside a closure or provider to ensure it is only invoked at Execution time.
+This function may not be invoked at [Gradle Configuration time](https://docs.gradle.org/current/userguide/build_lifecycle.html). Put it inside a closure or provider to ensure it is only invoked at Execution time.
 
-By default, this function resolve the `unifiedClasspath` configuration to supply a version, but you can always supply a different configuration if you want to:
+By default, this function supplies the version locked in `versions.lock`, but you can always supply a different configuration to resolve a version from instead:
 
 ```gradle
 task printGuavaVersion {
--- a/gradle/gradle-test-versions.yml
+++ b/gradle/gradle-test-versions.yml
@@ -1,3 +1,4 @@
 major-versions:
   8: 8.14.3
+  9: 9.3.1
 extra-versions: []
--- a/gradle/gradle-test-versions.yml
+++ b/gradle/gradle-test-versions.yml
@@ -1,4 +1,3 @@
 major-versions:
   8: 8.14.3
-  9: 9.3.1
 extra-versions: []
--- a/src/main/java/com/palantir/gradle/versions/GetVersionPlugin.java
+++ b/src/main/java/com/palantir/gradle/versions/GetVersionPlugin.java
@@ -29,8 +29,10 @@ import org.gradle.api.GradleException;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.artifacts.Configuration;
+import org.gradle.api.artifacts.DependencyConstraint;
 import org.gradle.api.artifacts.ModuleVersionIdentifier;
 import org.gradle.api.artifacts.result.ResolvedComponentResult;
+import org.gradle.api.provider.Provider;
 
 public final class GetVersionPlugin implements Plugin<Project> {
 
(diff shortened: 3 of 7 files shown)


FinlayRJW:
going the root of https://github.com/palantir/gradle-consistent-versions/pull/1667

Issue #1637, opened 2026-03-25
Migrate away from DependencyHandler.create(Project)


ghale:
# Summary

Gradle 9.6.0 will deprecate passing a Project object as dependency notation to DependencyHandler.create(). The recommended replacements are `DependencyHandler.project()` or `DependencyFactory.createProjectDependency()`. There are 3 call sites that need migration.

## Affected call sites

All three use the pattern (ProjectDependency) project.getDependencies().create(someProject):

1. [VersionsLockPlugin.java line 245](https://github.com/palantir/gradle-consistent-versions/blob/develop/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java#L245) — apply()
```
  ProjectDependency locksDependency =
          (ProjectDependency) project.getDependencies().create(project);
```

Suggested change:
```
  ProjectDependency locksDependency =
          (ProjectDependency) project.getDependencyFactory().createProjectDependency();
```

2. [VersionsLockPlugin.java line 504](https://github.com/palantir/gradle-consistent-versions/blob/develop/src/main/java/com/palantir/gradle/versions/VersionsLockPlugin.java#L504) — createDependencyOnProjectWithScope()
```
  ProjectDependency projectDependency =
          (ProjectDependency) project.getDependencies().create(project);
```

Suggested change:
```
  ProjectDependency projectDependency =
          (ProjectDependency) project.getDependencyFactory.createProjectDependency();
```

3. [VersionsPropsPlugin.java line 133](https://github.com/palantir/gradle-consistent-versions/blob/develop/src/main/java/com/palantir/gradle/versions/VersionsPropsPlugin.java#L133) — createDepOnRootConstraintsConfiguration()
```
  ProjectDependency projectDep =
          ((ProjectDependency) project.getDependencies().create(project.getRootProject()));
```

Suggested change:
```
  ProjectDependency projectDep =
          ((ProjectDependency) project.getDependencyFactory().createProjectDependency(project.getRootProject().getPath()));
```

PR #1667, opened 2026-06-02
Make `getVersion` resolve a per-project view of the `:unifiedClasspath`

FinlayRJW:
## Before this PR

`getVersion('group:name')` (from `GetVersionPlugin`) lets build scripts look up the version GCV settled on for a dependency. The no-configuration overloads worked by **resolving the root project's `unifiedClasspath`** to read the version back out.

On Gradle 9 this fails when `getVersion` is called from a subproject:

```
Caused by: org.gradle.api.internal.artifacts.configurations.DefaultConfiguration$IllegalResolutionException:
Resolution of the configuration ':unifiedClasspath' was attempted without an exclusive lock. This is unsafe and not allowed.
        at com.palantir.gradle.versions.GetVersionPlugin.getOptionalVersion(GetVersionPlugin.java:98)
        at com.palantir.gradle.versions.GetVersionPlugin.getVersion(GetVersionPlugin.java:83)
```

This was a warning on Gradle 8 and is now an error: a subproject may not resolve the root project's configuration. It commonly surfaces via `sls-packaging`'s `minimumVersion` usage inside `subprojects { ... }`, but affects any subproject `getVersion` call.

#1664 fixed the same error a different way, by **reading** the strict constraints off the root `gcvLocks` platform instead of resolving. That works, but still reads the root project's model from a subproject, and needs special handling for the `getVersion(project(...))` cases. This PR takes the isolated-projects-shaped route instead: each project resolves **its own** configuration.

## After this PR

==COMMIT_MSG==
Fix `getVersion(...)` failing with `Resolution of the configuration ':unifiedClasspath' was attempted without an exclusive lock` when called from a subproject on Gradle 9. Each project now resolves its own view of the unified classpath rather than the root project's configuration.
==COMMIT_MSG==

Each project resolves a **per-project view** of the unified dependency graph instead of reaching into the root's `unifiedClasspath`. The idea is a dependency-scope **configuration** that both the resolvable (lock-computing) and consumable views extend, so no configuration extends one of a different role.

- **`VersionsLockPlugin`** (root): collects every project's production + test deps into a new dependency-scope configuration, `unifiedClasspathDependencies`. Two views extend it: `unifiedClasspath` (resolvable) still computes the lock state — behaviour is unchanged — and a new `unifiedClasspathElements` (consumable) exposes the same graph to other projects under capability `gcv:unified-classpath:0`.
- **`GetVersionPlugin`** (applied to every project): registers a resolvable `gcvGetVersions` configuration that depends on `project(':')` requesting capability `gcv:unified-classpath:0`, and wires `getVersion(...)` to resolve that configuration.

```mermaid
flowchart TD
    subgraph root["root project"]
        dependencies["<b>unifiedClasspathDependencies</b><br/>(dependencies from every project's locked configurations)"]
        elements["<b>unifiedClasspathElements</b><br/>capability gcv:unified-classpath:0"]
        unified["<b>unifiedClasspath</b><br/>→ writeLocks / verifyLocks"]
        elements -->|extendsFrom| dependencies
        unified -->|extendsFrom| dependencies
    end
    subgraph child["every subproject (including root)"]
        gv["getVersion('group:name')"]
        getv["<b>gcvGetVersions</b>"]
        gv -->|resolves| getv
    end
    getv -->|"depends on, with capability gcv:unified-classpath:0"| elements

    classDef declarable fill:#3b82f6,stroke:#1d4ed8,color:#fff
    classDef resolvable fill:#a855f7,stroke:#7e22ce,color:#fff
    classDef consumable fill:#22c55e,stroke:#15803d,color:#fff
    classDef func fill:#e5e7eb,stroke:#9ca3af,color:#111

    class dependencies declarable
    class unified,getv resolvable
    class elements consumable
    class gv func

    style root fill:none,stroke:#bbb,color:#111
    style child fill:none,stroke:#bbb,color:#111
```

Because `getVersion` resolves the calling project's own `gcvGetVersions` configuration — which, via the consumable view, sees exactly the same dependencies as before — the version returned is identical, and it never resolves a cross-project configuration, so the exclusive-lock error is gone. The one behaviour change is the explicit-configuration overload `getVersion('group:name', configuration)`: since each project now resolves its own configuration, the supplied configuration must belong to the calling project, and `getVersion` fails fast with a clear error otherwise.


README.md
```diff
@@ -208,9 +208,9 @@ task printGuavaVersion {
 }
 ```
 
-This function may not be invoked at [Gradle Configuration time](https://docs.gradle.org/current/userguide/build_lifecycle.html) as it involves resolving dependencies.  Put it inside a closure or provider to ensure it is only invoked at Execution time.
+This function may not be invoked at [Gradle Configuration time](https://docs.gradle.org/current/userguide/build_lifecycle.html) as it involves resolving dependencies. Put it inside a closure or provider to ensure it is only invoked at Execution time.
 
-By default, this function resolve the `unifiedClasspath` configuration to supply a version, but you can always supply a different configuration if you want to:
+By default, this function resolves the unified classpath to supply a version, but you can always supply a different configuration to resolve a version from instead:
```
felixdesouza: > unified classpath

isn't an existing concept in gradle, so what does it refer to when reading this?

`unifiedClasspath` as a configuration _is_ something that exists in the README and is technically accurate. I would prefer you stick to that nomenclature. Understandably with this change, it's not strictly true, it's more like "this function will eventually resolve the `unifiedClasspath` configuration located in the root project to supply a version.

> but you can always supply a different configuration to resolve a version from instead

Probably worth noting in the README that whatever configuration supplied needs to live in the project that `getVersion` is being called from. Not sure what error message you get when you don't, but it could be nice to make that clearer if it's not obvious already - i.e make a better error message if it's unclear.

src/test/java/com/palantir/gradle/versions/GetVersionPluginIntegrationTest.java
```diff
@@ -71,6 +72,22 @@ void get_version_is_callable_from_groovy_with_two_string_args_and_configuration_
...
+            tasks.register('callGetVersion') {
+                doLast {
+                    getVersion('com.google.guava', 'guava')
+                }
+            }
+            """);
+        InvocationResult result = gradle.withArgs(":child:callGetVersion").buildsWithFailure();
+        assertThat(result)
+                .output()
+                .contains("Unable to find 'com.google.guava:guava' in configuration ':child:gcvGetVersions'");
```
felixdesouza: this seems like a bad test? the assertion does not jibe with the name of the test, how do I know that it get version is actually just not callable anymore but has a nicer error message?

src/test/java/com/palantir/gradle/versions/ConsistentVersionsPluginIntegrationTest.java
```diff
@@ -179,6 +179,57 @@ void get_version_function_works_even_when_writing_locks(GradleInvoker gradle, Ro
...
+
+        InvocationResult result =
+                gradle.withArgs(":child:demo", "--write-locks").buildsSuccessfully();
+        assertThat(result).output().contains("demo=1.7.25");
+    }
+
+    @Test
+    @DisabledConfigurationCache(
+            "Cannot reference a Gradle script object from a Groovy closure as these are not supported with the"
+                    + " configuration cache")
```
felixdesouza, line 212: That is interesting, this pretty much puts a *lot* of repos out of contention for doing configuration cache in the future

felixdesouza: not for this PR, but I wonder whether there are other ways to allow this function to be called but not as a groovy closure

src/test/java/com/palantir/gradle/versions/ConsistentVersionsPluginIntegrationTest.java
```diff
@@ -179,6 +179,57 @@ void get_version_function_works_even_when_writing_locks(GradleInvoker gradle, Ro
         assertThat(result).output().contains("demo=1.7.25");
     }
 
+    @Test
+    @DisabledConfigurationCache(
+            "Cannot reference a Gradle script object from a Groovy closure as these are not supported with the"
+                    + " configuration cache")
+    void get_version_function_works_from_subproject(GradleInvoker gradle, RootProject rootProject, SubProject child) {
+        rootProject.propertiesFile("versions.props").setProperty("org.slf4j:*", "1.7.25");
```
felixdesouza, line 191: huh, inadvertent use of the properties file, don't think I would have paired them together, but that _is_ interesting

src/test/java/com/palantir/gradle/versions/ConsistentVersionsPluginIntegrationTest.java
```diff
@@ -179,6 +179,57 @@ void get_version_function_works_even_when_writing_locks(GradleInvoker gradle, Ro
         assertThat(result).output().contains("demo=1.7.25");
     }
 
+    @Test
+    @DisabledConfigurationCache(
+            "Cannot reference a Gradle script object from a Groovy closure as these are not supported with the"
+                    + " configuration cache")
+    void get_version_function_works_from_subproject(GradleInvoker gradle, RootProject rootProject, SubProject child) {
```
felixdesouza, line 190: I think another test that would be interesting is getting a version for something that is *not* in your project or dependency list.

* This test gets the version for something it directly depends on
* The below test gets the version for the project it depends on - through its group

Unless of course this test already exists, will need to check

src/main/java/com/palantir/gradle/versions/GetVersionPlugin.java
```diff
@@ -18,70 +18,50 @@
...
 
-    private static String getVersion(Project project, String group, String name, Configuration configuration) {
-        return getOptionalVersion(project, group, name, configuration)
-                .orElseThrow(() -> notFound(group, name, configuration));
+        rootProject
+                .getAllprojects()
+                .forEach(project -> project.getPluginManager().apply(GetVersionProjectPlugin.class));
     }
 
     static Optional<String> getOptionalVersion(
```
felixdesouza, line 131: move this to where it's used, i.e. `GetVersionProjectPlugin`, it's not used within this class at all anymore

src/main/java/com/palantir/gradle/versions/GetVersionProjectPlugin.java
```diff
@@ -0,0 +1,126 @@
...
+                    configuration
+                            .getAttributes()
+                            .attribute(
+                                    GcvBuildPath.ATTRIBUTE,
+                                    getObjects()
+                                            .newInstance(GcvAttributes.class)
+                                            .buildPath());
+
+                    ProjectDependency rootDependency =
+                            (ProjectDependency) getDependencies().project(ImmutableMap.of("path", ":"));
```
felixdesouza: look at: https://github.com/palantir/gradle-consistent-versions/issues/1637

let's use the right invocation

felixdesouza: oh, that's a bit difficult, considering we're gradle 8, might potentially have to do a reflection piece potentially, not sure when it'll get deleted, perhaps gradle 10 idk

src/main/java/com/palantir/gradle/versions/GetVersionProjectPlugin.java
```diff
@@ -0,0 +1,126 @@
...
+                    configuration.setCanBeResolved(true);
+                    configuration
+                            .getAttributes()
+                            .attribute(VersionsLockPlugin.GCV_USAGE_ATTRIBUTE, GcvUsage.GCV_SOURCE);
+                    configuration
+                            .getAttributes()
+                            .attribute(
+                                    GcvBuildPath.ATTRIBUTE,
+                                    getObjects()
+                                            .newInstance(GcvAttributes.class)
```
felixdesouza: move this to `@Nested`, so the nesting here isn't insane

src/main/java/com/palantir/gradle/versions/GetVersionProjectPlugin.java
felixdesouza: can we also do `project.getRootProject().getPath()` ?

README.md
```diff
@@ -208,9 +208,9 @@ task printGuavaVersion {
 }
 ```
 
-This function may not be invoked at [Gradle Configuration time](https://docs.gradle.org/current/userguide/build_lifecycle.html) as it involves resolving dependencies.  Put it inside a closure or provider to ensure it is only invoked at Execution time.
+This function may not be invoked at [Gradle Configuration time](https://docs.gradle.org/current/userguide/build_lifecycle.html) as it involves resolving dependencies. Put it inside a closure or provider to ensure it is only invoked at Execution time.
```
felixdesouza, line 211: Perhaps not in the README, but in some sort of ADR or something, I think it'd be good to describe how the `GetVersionsPlugin` works, i.e. basically your PR description, but somewhere checked into the repo or something. Not blocking however.

felixdesouza (commented):
Directionally, I think this makes a lot of sense. I would probably try to run it on some of the gnarlier internal repos to see if there are any edge cases we're missing.

src/test/java/com/palantir/gradle/versions/ConsistentVersionsPluginIntegrationTest.java
FinlayRJW: is there anything wrong with using it this way? I've been doing so in a few places https://pl.ntr/2BI can always excavate / add an errorprone if we think this is a bad pattern

src/main/java/com/palantir/gradle/versions/GetVersionProjectPlugin.java
FinlayRJW: #1637 says "recommended replacements are DependencyHandler.project() or DependencyFactory.createProjectDependency()"  `createProjectDependency` is gradle 9 only but we are already using `getDependencies().project` so I think 

```java
ProjectDependency rootDependency = (ProjectDependency) getDependencies()
                            .project(Map.of("path", project.getRootProject().getPath()));
```

should be fine - we can always migrate to `createProjectDependency` once we are dropping gradle 8 support?

FinlayRJW: using `project.getRootProject().getPath()` in [b668450](https://github.com/palantir/gradle-consistent-versions/pull/1667/commits/b668450bb8656d459b42f732a9db32306730861b)

src/test/java/com/palantir/gradle/versions/GetVersionPluginIntegrationTest.java
FinlayRJW: agreed this was how I originally repo'd the issue but the tests in `ConsistentVersionsPluginIntegrationTest` supersede it. Removed in [aad524d](https://github.com/palantir/gradle-consistent-versions/pull/1667/commits/aad524d2028dbc04751597adc945801616076306)

src/test/java/com/palantir/gradle/versions/ConsistentVersionsPluginIntegrationTest.java
FinlayRJW: yeah think this could cause some pretty gnarly issues, noted as a CC FLUP

src/test/java/com/palantir/gradle/versions/ConsistentVersionsPluginIntegrationTest.java
FinlayRJW: I don't think we quite have a test that covers this. I'v added `get_version_function_works_for_a_dependency_not_declared_by_the_project` to `ConsistentVersionsPluginIntegrationTest.java` in [9e0ac95](https://github.com/palantir/gradle-consistent-versions/pull/1667/commits/9e0ac959e1ec538ede1b321834e35dc167a16de7):

 a `consumer` subproject that declares no dependencies at all  successfully resolves `getVersion('org.slf4j:slf4j-api')` to `1.7.25`, where slf4j is only declared by the root project. 

is this what you where thinking?

src/main/java/com/palantir/gradle/versions/GetVersionPlugin.java
FinlayRJW: moved in [853c372](https://github.com/palantir/gradle-consistent-versions/pull/1667/commits/853c372c6f5eb854482eb1db3fe3cf645fc25ebc)

src/main/java/com/palantir/gradle/versions/GetVersionProjectPlugin.java
FinlayRJW: nice!, added `configureGcvSourceAttributes` as well as shared with `VersionsLockPlugin` in [663d74b](https://github.com/palantir/gradle-consistent-versions/pull/1667/commits/663d74b08576d6fbde5503aea17606167ed87a3c)

README.md
FinlayRJW: updated to:

"By default, this function will eventually resolve the `unifiedClasspath` configuration located in the root project to supply a version. You can also supply a different configuration to resolve a version from instead. Note that any configuration you supply must live in the project that `getVersion` is being called from:" in [c3113d9](https://github.com/palantir/gradle-consistent-versions/pull/1667/commits/c3113d9f9036989e8ef3d0189dafa511398569fd)

README.md
FinlayRJW: added in `docs/adr/0001-getversion-per-project-resolution.md`

FinlayRJW:
> I would probably try to run it on some of the gnarlier internal repos to see if there are any edge cases we're missing.

tested internally on three large repos and saw no issues see 🧵 [here for more details](https://pl.ntr/2BJ)

docs/adr/0001-getversion-per-project-resolution.md
```diff
@@ -0,0 +1,85 @@
+# 1. `getVersion` resolves a per-project view of the unified classpath
```
felixdesouza: please turn on line wrapping in your IDE, and have things be single line.

It makes editing and future diffs incredibly annoying to read without this.

docs/adr/0001-getversion-per-project-resolution.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Context
+
+`getVersion('group:name')` (from `GetVersionPlugin`) lets build scripts look up the
+version GCV settled on for a dependency. The no-configuration overloads originally
+worked by resolving the **root** project's `unifiedClasspath` configuration to read the
```
felixdesouza: you haven't introduced the overloads, let's not talk about them, considering you've been very specific in the form that you're referring to.


```suggestion
version GCV settled on for a dependency. Originally, this
worked by resolving the **root** project's `unifiedClasspath` configuration to read the
```

docs/adr/0001-getversion-per-project-resolution.md
```diff
@@ -0,0 +1,85 @@
...
+version back out, regardless of which project `getVersion` was called from.
+
+On Gradle 9 this fails when `getVersion` is called from a subproject:
```
felixdesouza: it's weird to put `getVersion` here because now you have overloads to think about, I would just say


```suggestion
On Gradle 9 this fails when it is called from a subproject:
```

docs/adr/0001-getversion-per-project-resolution.md
```diff
@@ -0,0 +1,85 @@
...
+Caused by: org.gradle.api.internal.artifacts.configurations.DefaultConfiguration$IllegalResolutionException:
+Resolution of the configuration ':unifiedClasspath' was attempted without an exclusive lock. This is unsafe and not allowed.
+        at com.palantir.gradle.versions.GetVersionPlugin.getOptionalVersion(GetVersionPlugin.java:98)
+        at com.palantir.gradle.versions.GetVersionPlugin.getVersion(GetVersionPlugin.java:83)
+```
+
+This was a warning on Gradle 8 and is now an error: a subproject may not resolve the
+root project's configuration. It commonly surfaces via `sls-packaging`'s
+`minimumVersion` usage inside `subprojects { ... }`, but affects any subproject
+`getVersion` call.
```
felixdesouza: I would put an example of the sls-packaging usage, it is not obvious what you're referring to.

docs/adr/0001-getversion-per-project-resolution.md
```diff
@@ -0,0 +1,85 @@
...
+
+```
+Caused by: org.gradle.api.internal.artifacts.configurations.DefaultConfiguration$IllegalResolutionException:
+Resolution of the configuration ':unifiedClasspath' was attempted without an exclusive lock. This is unsafe and not allowed.
+        at com.palantir.gradle.versions.GetVersionPlugin.getOptionalVersion(GetVersionPlugin.java:98)
+        at com.palantir.gradle.versions.GetVersionPlugin.getVersion(GetVersionPlugin.java:83)
+```
+
+This was a warning on Gradle 8 and is now an error: a subproject may not resolve the
+root project's configuration. It commonly surfaces via `sls-packaging`'s
```
felixdesouza: > a subproject may not resolve the root project's configuration

is this supposed to be a reference to an error? or something else

docs/adr/0001-getversion-per-project-resolution.md
```diff
@@ -0,0 +1,85 @@
...
+`minimumVersion` usage inside `subprojects { ... }`, but affects any subproject
+`getVersion` call.
+
+### Approaches considered
+
+- **Read the `versions.lock` file directly.** All the information is there, but reading
+  the lockfile from a subproject feels against the intent of GCV.
```
felixdesouza: I think you need to clarify a bit more here. `GetVersionPlugin` is _part_ of GCV.

docs/adr/0001-getversion-per-project-resolution.md
```diff
@@ -0,0 +1,85 @@
...
+- **Read the `gcvLocks` constraints instead of resolving.** The "without an exclusive
+  lock" error only applies to *resolution* — reading a configuration's declared
+  `DependencyConstraint`s is fine, so the no-config lookup could read the strict
```
felixdesouza: can we drop the `no-config` bits throughout this doc pls

docs/adr/0001-getversion-per-project-resolution.md
```diff
@@ -0,0 +1,85 @@
...
+root project's configuration. It commonly surfaces via `sls-packaging`'s
+`minimumVersion` usage inside `subprojects { ... }`, but affects any subproject
+`getVersion` call.
+
+### Approaches considered
+
+- **Read the `versions.lock` file directly.** All the information is there, but reading
+  the lockfile from a subproject feels against the intent of GCV.
+- **Read the `gcvLocks` constraints instead of resolving.** The "without an exclusive
+  lock" error only applies to *resolution* — reading a configuration's declared
```
felixdesouza:
```suggestion
- **Read the `gcvLocks` constraints instead of resolving.** The above error only applies to *resolution* — reading a configuration's declared
```

docs/adr/0001-getversion-per-project-resolution.md
```diff
@@ -0,0 +1,85 @@
...
+  configuration instead of reaching into the root's `unifiedClasspath`. This is the
+  isolated-projects-shaped route and is the only one that keeps all `getVersion`
+  overloads working without special-casing.
+
+## Decision
+
+Each project resolves a per-project view of the unified dependency graph instead of
+reaching into the root's `unifiedClasspath`.
+
+The trick is a **dependency-scope bucket** that both the resolvable (lock-computing) and
```
felixdesouza: let's not introduce new nomenclature for things that have existing names. "bucket" in this case should be "configuration"

felixdesouza: I'm also not a fan of "trick". This is how we're supposed to do stuff with configurations, we were just doing it wrong before/gradle moved on to a different system that we're adopting to now.

docs/adr/0001-getversion-per-project-resolution.md
```diff
@@ -0,0 +1,85 @@
...
+consumable (per-project) views extend, so no configuration extends one of a different
```
felixdesouza: this is not true? you can't extend configurations outside of your project.

felixdesouza: https://docs.gradle.org/current/userguide/upgrading_version_8.html#extending_configurations_in_same_project

docs/adr/0001-getversion-per-project-resolution.md
```diff
@@ -0,0 +1,85 @@
...
+role (Gradle 9 warns on a consumable extending a resolvable, and vice versa).
+
+- **`VersionsLockPlugin`** (root): collects every project's production + test deps into a
```
felixdesouza: deps -> dependencies

docs/adr/0001-getversion-per-project-resolution.md
```diff
@@ -0,0 +1,85 @@
...
+  new dependency-scope bucket, `unifiedClasspathDependencies`. `unifiedClasspath`
+  (resolvable) now just `extendsFrom` the bucket and still computes the lock state —
+  behaviour is unchanged, the bucket is purely an intermediate.
+- **`GetVersionPlugin`** (root): adds `gcvGetVersionElements`, a *consumable* view that
+  also extends the bucket, carrying capability `gcv:get-versions:0` and the `GCV_SOURCE`
+  usage. Applies `GetVersionProjectPlugin` to every project.
+- **`GetVersionProjectPlugin`** (new, per-project): registers a resolvable
+  `gcvGetVersions` that depends on `project(':')` requesting that capability, and wires
+  the `getVersion(...)` extension to resolve that configuration.
```
felixdesouza:
```suggestion
- **`GetVersionProjectPlugin`** (new, per-project): registers a resolvable
  `gcvGetVersions` configuration that depends on `project(':')` requesting the same capability, and wires
  the `getVersion(...)` ? to resolve that configuration.
```

`getVersion` is not an extension?

docs/adr/0001-getversion-per-project-resolution.md
```diff
@@ -0,0 +1,85 @@
...
+The trick is a **dependency-scope bucket** that both the resolvable (lock-computing) and
+consumable (per-project) views extend, so no configuration extends one of a different
+role (Gradle 9 warns on a consumable extending a resolvable, and vice versa).
+
+- **`VersionsLockPlugin`** (root): collects every project's production + test deps into a
+  new dependency-scope bucket, `unifiedClasspathDependencies`. `unifiedClasspath`
+  (resolvable) now just `extendsFrom` the bucket and still computes the lock state —
+  behaviour is unchanged, the bucket is purely an intermediate.
+- **`GetVersionPlugin`** (root): adds `gcvGetVersionElements`, a *consumable* view that
+  also extends the bucket, carrying capability `gcv:get-versions:0` and the `GCV_SOURCE`
```
felixdesouza: don't think the `GCV_SOURCE` is worth mentioning

docs/adr/0001-getversion-per-project-resolution.md
```diff
@@ -0,0 +1,85 @@
...
+reaching into the root's `unifiedClasspath`.
+
+The trick is a **dependency-scope bucket** that both the resolvable (lock-computing) and
+consumable (per-project) views extend, so no configuration extends one of a different
+role (Gradle 9 warns on a consumable extending a resolvable, and vice versa).
+
+- **`VersionsLockPlugin`** (root): collects every project's production + test deps into a
+  new dependency-scope bucket, `unifiedClasspathDependencies`. `unifiedClasspath`
+  (resolvable) now just `extendsFrom` the bucket and still computes the lock state —
+  behaviour is unchanged, the bucket is purely an intermediate.
```
felixdesouza: The phrasing here still feels like we're describing a trick, when in actual fact it's what we're supposed to do.

> the bucket is purely an intermediate

This is essentially fluff, it's the entire purpose for dependency scope configurations, so either should be assumed or described as such.

docs/adr/0001-getversion-per-project-resolution.md
```diff
@@ -0,0 +1,85 @@
...
+    style root fill:none,stroke:#bbb,color:#111
+    style child fill:none,stroke:#bbb,color:#111
+```
+
+## Consequences
+
+Because `getVersion` resolves the unified graph in the *calling* project, the result
+still carries both locked external versions and in-build project versions, and it never
+resolves a cross-project configuration — so the exclusive-lock error is gone and the
+`getVersion('group:name', configuration)` / `getVersion(project(...))` cases keep working
```
felixdesouza: > resolves the unified graph in the *calling* project

* what is the unified graph?
* if you're referring to the `unifiedClasspath`, it's a little weird to say it's resolved in the *calling* project, the whole premise is that you're not resolving that specific configuration in the calling project.

> `getVersion('group:name', configuration)` / `getVersion(project(...))` cases keep working

This is the first time you're mentioning these overloads, why do these matter and why is it a corollary to the work you've just done. Genuinely asking because I don't see how it follows

docs/adr/0001-getversion-per-project-resolution.md
```diff
@@ -0,0 +1,85 @@
...
+- **`GetVersionPlugin`** (root): adds `gcvGetVersionElements`, a *consumable* view that
+  also extends the bucket, carrying capability `gcv:get-versions:0` and the `GCV_SOURCE`
+  usage. Applies `GetVersionProjectPlugin` to every project.
+- **`GetVersionProjectPlugin`** (new, per-project): registers a resolvable
+  `gcvGetVersions` that depends on `project(':')` requesting that capability, and wires
+  the `getVersion(...)` extension to resolve that configuration.
+
+```mermaid
+flowchart TD
+    subgraph root["root project"]
```
felixdesouza, line 47: * > every project's prod + test deps
   
    This should be: 

     > dependencies from every project's locked configurations

* your arrows are the wrong way in a number of places:
  * `gcvGetVersionElements` and `unifiedClasspath` both `extendsFrom` `unifiedClasspathDependencies`, and yet, the arrow seems to suggest the other way.
  * I would reverse the arrow between `getVersion` and `gcvGetVersions`, and add a label along the lines of `resolves`
  * from `gcvGetVersions` -> `gcvGetVersionElements`, I would instead say
     > depends on with capability gcv:get-versions:0
* > :child (and every project)
  
  Just say:

  > every subproject (including root)

* drop the bucket nomenclature everywhere, these are configurations not buckets! 
* It would be good to colour code the boxes, i.e. what's dependency scope, what's resolvable, what's consumable, what's a just a normal function. See the official gradle [docs](https://docs.gradle.org/current/userguide/java_library_plugin.html#resolvable_configurations) for the java library plugin that describes this.

docs/adr/0001-getversion-per-project-resolution.md
```diff
@@ -0,0 +1,85 @@
...
+
+The trick is a **dependency-scope bucket** that both the resolvable (lock-computing) and
+consumable (per-project) views extend, so no configuration extends one of a different
+role (Gradle 9 warns on a consumable extending a resolvable, and vice versa).
+
+- **`VersionsLockPlugin`** (root): collects every project's production + test deps into a
+  new dependency-scope bucket, `unifiedClasspathDependencies`. `unifiedClasspath`
+  (resolvable) now just `extendsFrom` the bucket and still computes the lock state —
+  behaviour is unchanged, the bucket is purely an intermediate.
+- **`GetVersionPlugin`** (root): adds `gcvGetVersionElements`, a *consumable* view that
```
felixdesouza: reading this more, I don't think it makes sense for `GetVersionPlugin` to add this? 

Typically, what I've seen is that you have:
* name of thing that was resolvable e.g. `runtimeClasspath`
* name of thing that was the dependency scope `implementation`
* the consumable version of the thing being suffixed with `Elements` e.g. `runtimeElements`

I think a similar approach makes sense here. i.e. you have:

* `unifiedClasspathDependencies` for the declarable configuartion
* `unifiedClasspath` for the resolvable - we can't change this for obvious reasons
* `unifiedClasspathElements` for the consumable.

The idea is that you want to have a consumable view of the `unifiedClasspath`, it's not restricted to just the `getVersion`s workflow, it just so happens to be the initial and only consumer (for now).

src/main/java/com/palantir/gradle/versions/GetVersionProjectPlugin.java
```diff
@@ -0,0 +1,149 @@
...
+                .orElseThrow(() -> notFound(group, name, configuration));
+    }
+
+    static Optional<String> getOptionalVersion(
+            Project project, String group, String name, Configuration configuration) {
+        if (GradleWorkarounds.isConfiguring(project.getState())) {
+            throw new GradleException(String.format(
+                    "Not allowed to call gradle-consistent-versions's getVersion(\"%s\", \"%s\", "
+                            + "configurations.%s) "
+                            + "at configuration time",
```
felixdesouza: let's make this a multline string whilst you're here. I think it's also gone a little wonky with the string appending

src/main/java/com/palantir/gradle/versions/GetVersionProjectPlugin.java
```diff
@@ -0,0 +1,149 @@
...
+
+    private static GradleException notFound(String group, String name, Configuration configuration) {
+        String actual = configuration.getIncoming().getResolutionResult().getAllComponents().stream()
+                .map(ResolvedComponentResult::getModuleVersion)
+                .map(mvi -> String.format("\t- %s:%s:%s", mvi.getGroup(), mvi.getName(), mvi.getVersion()))
+                .collect(Collectors.joining("\n"));
+        return new GradleException(String.format(
+                "Unable to find '%s:%s' in %s. This may happen if you specify the version in versions.props"
+                        + " but do not have a dependency in the configuration. The configuration contained:\n"
+                        + "%s",
```
felixdesouza: same thign here re multiline

README.md
```diff
@@ -208,9 +208,9 @@ task printGuavaVersion {
...
 
-By default, this function resolve the `unifiedClasspath` configuration to supply a version, but you can always supply a different configuration if you want to:
+By default, this function will eventually resolve the `unifiedClasspath` configuration located in the root project to supply a version ([how this works](docs/adr/0001-getversion-per-project-resolution.md)). You can also supply a different configuration to resolve a version from instead. Note that any configuration you supply must live in the project that `getVersion` is being called from:
```
felixdesouza: let's put:

> that any configuration you supply must live in the project that `getVersion` is being called from

in a callout.

The colon no longer follows the content. It should be after:

> You can also supply a different configuration to resolve a version from instead:

Where it is now has nothing to do with the following content.

docs/adr/0001-getversion-per-project-resolution.md
FinlayRJW: good idea! then we wouldn't need two `GetVersionPlugin`s - fixed in a0fa53e2644c06ac261fb74e641b0511a75fba85

src/main/java/com/palantir/gradle/versions/GetVersionProjectPlugin.java
FinlayRJW: fixed in a0fa53e2644c06ac261fb74e641b0511a75fba85

src/main/java/com/palantir/gradle/versions/GetVersionProjectPlugin.java
FinlayRJW: fixed in a0fa53e2644c06ac261fb74e641b0511a75fba85

docs/adr/0001-getversion-per-project-resolution.md
FinlayRJW: fixed in 532d8923a46ec0fea09082ea723aa8a0d78353b9

docs/adr/0001-getversion-per-project-resolution.md
FinlayRJW: example added in 1b55dd3f37d5b6b5472343fed9d9250687a108e8

docs/adr/0001-getversion-per-project-resolution.md
FinlayRJW: This was supposed to reference the above error but honestly doesn't add anything useful so removed in 1b55dd3f37d5b6b5472343fed9d9250687a108e8

docs/adr/0001-getversion-per-project-resolution.md
FinlayRJW: 👍  fixed in 5b4be03ef8bd115122300ffef342c72597cb7b5e

docs/adr/0001-getversion-per-project-resolution.md
FinlayRJW: nice have extended in 5b4be03ef8bd115122300ffef342c72597cb7b5e also noted that this approach doesnt work as the lock only contains external modules

docs/adr/0001-getversion-per-project-resolution.md
FinlayRJW: removed the bucked reference 216df7d7ec08693b81ea430e1731c572c6ff2319

FinlayRJW: fix the "trick" reference here 3d303e2bfdb90941e659a0255d67cb2d57546826

docs/adr/0001-getversion-per-project-resolution.md
FinlayRJW: whoops I meant root-project, fixed and made clearer (i hope) in b6be120b66d6ee7d7dbfdcdc2bd00fe0d5b8c0b9

docs/adr/0001-getversion-per-project-resolution.md
FinlayRJW: :+1 fixed in c9b9a1b5aabbe4bc08c1eacd13aba7fc1416243c

docs/adr/0001-getversion-per-project-resolution.md
FinlayRJW: fixed in c9b9a1b5aabbe4bc08c1eacd13aba7fc1416243c

docs/adr/0001-getversion-per-project-resolution.md
FinlayRJW: sorry I have such a habit of shortening thing! fixed in b2411a72d361f998cfe846e691bce3cf94b007fc

docs/adr/0001-getversion-per-project-resolution.md
FinlayRJW: I was referencing all the dependencies that are stored in the new dependency scoped configuration  that the consumable extends as didn't want to use `unifiedClasspath` when thats not what we are actually resolving - will make this more explicit

> first time you're mentioning these overloads

I wanted to make it clear that these had not changed (which isn't true anymore 9ed01041f682c8051b7889394a701d4972745143) will update to reflect how it actually is!

FinlayRJW: hopefully I made this clearer here 9398402cfdfb646056878b788d0e482e64f21c72

docs/adr/0001-getversion-per-project-resolution.md
FinlayRJW: between f3516b9d0b471ad4a8e0b35c0a5b567ac05a2e4e and 556d4e3e12ea42773f0ce74171dfa6637cd52590 should have it better

README.md
FinlayRJW: fixed in 93318d4b89859b7f09e8519decc253fe350c6ce7

docs/adr/0001-getversion-per-project-resolution.md
```diff
@@ -0,0 +1,79 @@
...
+        minimumVersion = getVersion('com.palantir.group', 'my-service-module-api')
+        maximumVersion = "1.x.x"
+    }
+}
+```
+
+### Approaches considered
+
+- **Read the `versions.lock` file directly.** The `getVersion('group:name')` lookup could parse the resolved version straight out of `versions.lock`. But `GetVersionPlugin` is as part of GCV itself, it has direct access to the very same resolved dependency model GCV uses to *write* `versions.lock`, so re-parsing its own serialised output would be backwards. It also wouldn't work: the lockfile only records *external* modules, so it couldn't serve a `getVersion` lookup that resolves to another project in the same build.
+- **Read the `gcvLocks` constraints instead of resolving.** The above error only applies to *resolution* — reading a configuration's declared `DependencyConstraint`s is fine, so the lookup could read the strict constraints off the root `gcvLocks` platform instead. This works for the common case but still reads the root project's model from a subproject, it needs special-casing for the `getVersion(project(...))` case.
```
felixdesouza: > it needs special-casing for the `getVersion(project(...))` case.

and 

> is the only one that keeps all `getVersion` overloads working without special-casing.

It's not clear what special casing you're referring to here.

docs/adr/0001-getversion-per-project-resolution.md
```diff
@@ -0,0 +1,79 @@
...
+- **Per-project view of the unified graph (chosen).** Each project resolves its *own* configuration instead of reaching into the root's `unifiedClasspath`. This is the isolated-projects-shaped route and is the only one that keeps all `getVersion` overloads working without special-casing.
+
+## Decision
+
+Each project resolves a per-project view of the unified dependency graph instead of reaching into the root's `unifiedClasspath`.
+
+This follows Gradle's intended separation of configuration roles. On the root project, a **dependency-scope configuration** holds the declared dependencies, and both the resolvable view that computes the locks and the consumable view that exposes them extend it. Each project then gets its own resolvable configuration that *consumes* that root view through a project dependency requesting its capability.
```
felixdesouza: `view` -> `configuration`, they are configurations!

docs/adr/0001-getversion-per-project-resolution.md
```diff
@@ -0,0 +1,79 @@
...
+
+- **`VersionsLockPlugin`** (root): collects every project's production + test dependencies into a new dependency-scope configuration, `unifiedClasspathDependencies`. Two views extend it: `unifiedClasspath` (resolvable) still computes the lock state — behaviour is unchanged — and a new `unifiedClasspathElements` (consumable) exposes the same graph to other projects under capability `gcv:unified-classpath:0`.
+- **`GetVersionPlugin`** (applied to every project): registers a resolvable `gcvGetVersions` configuration that depends on `project(':')` requesting capability `gcv:unified-classpath:0`, and wires `getVersion(...)` to resolve that configuration.
```
felixdesouza: > to resolve that configuration

configuration is overloaded here. 

I would say something along the lines of:

> * registers a resolvable `gcvGetVersions` configuration that effectively points to `unifiedClasspathElements` from above by depending on the root project and requiring the same capability: `gcv:unified-classpath:0`. 
> * `getVersion("some:dependency")` is then configured to resolve against `gcvGetVersions` in the _same_ instead of the `unifiedClasspath` configuration that resides in the root project.

docs/adr/0001-getversion-per-project-resolution.md
```diff
@@ -0,0 +1,79 @@
...
+
+    style root fill:none,stroke:#bbb,color:#111
+    style child fill:none,stroke:#bbb,color:#111
+```
+
+Box colours by configuration role: 🟦 **blue** = dependency-scope, 🟪 **purple** = resolvable, 🟩 **green** = consumable, ⬜ **grey** = plain function.
+
+## Consequences
+
+`getVersion` now resolves the calling project's own `gcvGetVersions` configuration. Via the capability-selected consumable view, that configuration sees exactly the same set of dependencies collected into `unifiedClasspathDependencies`, so the version returned is identical to before. The only difference is that the resolution happens in the calling project rather than reaching into another project's configuration, so the exclusive-lock error is gone.
```
felixdesouza: same thing here, please drop this view nomenclature

docs/adr/0001-getversion-per-project-resolution.md
felixdesouza: nice!

README.md
```diff
@@ -221,6 +221,9 @@ task printGuavaVersion {
 }
 ```
 
+> [!NOTE]
```
felixdesouza: let's make this a `WARNING` and make `must` bold

README.md
```diff
@@ -221,6 +221,9 @@ task printGuavaVersion {
...
+> Any configuration you supply must live in the project that `getVersion` is being called from.
```
felixdesouza:
```suggestion
> [!WARNING]
> Any configuration you supply **must** live in the project that `getVersion` is being called from.
```

docs/adr/0001-getversion-per-project-resolution.md
FinlayRJW: [fe80fae](https://github.com/palantir/gradle-consistent-versions/pull/1667/commits/fe80fae4f459a709173115ef3f0c3a5f00c0448a) fixed here

docs/adr/0001-getversion-per-project-resolution.md
FinlayRJW: fixed in [03f9599](https://github.com/palantir/gradle-consistent-versions/pull/1667/commits/03f9599211f0ba6c1f7020fcd0d16e86081f060d)

src/main/java/com/palantir/gradle/versions/GetVersionPlugin.java
```diff
@@ -80,18 +102,30 @@ public String doCall(String group, String name, Configuration configuration) {
...
+
     static Optional<String> getOptionalVersion(
             Project project, String group, String name, Configuration configuration) {
         if (GradleWorkarounds.isConfiguring(project.getState())) {
-            throw new GradleException(String.format(
-                    "Not allowed to call gradle-consistent-versions's getVersion(\"%s\", \"%s\", "
-                            + "configurations.%s) "
-                            + "at configuration time",
-                    group, name, configuration.getName()));
+            throw new GradleException(String.format("""
```
crogoz, line 134: might be useful to add a test for this as well (to make sure GradleWorkarounds.isConfiguring works correclty in this new setup)

docs/adr/0001-getversion-per-project-resolution.md
```diff
@@ -0,0 +1,81 @@
...
+        productName = "my-service"
+        minimumVersion = getVersion('com.palantir.group', 'my-service-module-api')
+        maximumVersion = "1.x.x"
+    }
+}
+```
+
+### Approaches considered
+
+- **Read the `versions.lock` file directly.** The `getVersion('group:name')` lookup could parse the resolved version straight out of `versions.lock`. But `GetVersionPlugin` is as part of GCV itself, it has direct access to the very same resolved dependency model GCV uses to *write* `versions.lock`, so re-parsing its own serialised output would be backwards. It also wouldn't work: the lockfile only records *external* modules, so it couldn't serve a `getVersion` lookup that resolves to another project in the same build.
```
crogoz, line 30: just flagging - gaia seems to do something similar to this one: https://github.palantir.build/acme/gaia/blob/8bc77e95f132498adce1c55b284ec4c8685dc411/build.gradle#L97. I am not sure if this would still work with the changes in this PR.

FinlayRJW: I think this should still work with this PR as they are overriding the `getVersion` for there own

FinlayRJW: I don't think this will make GCV CoD compatible so think they will still need that workaround

src/main/java/com/palantir/gradle/versions/GetVersionPlugin.java
FinlayRJW: good idea added in 871546488459de0cf1104410c86514d2da77ed88

src/main/java/com/palantir/gradle/versions/GetVersionPlugin.java
```diff
@@ -80,18 +102,30 @@ public String doCall(String group, String name, Configuration configuration) {
...
+    private static void checkConfigurationBelongsToProject(Project project, Configuration configuration) {
+        Configuration ownConfiguration = project.getConfigurations().findByName(configuration.getName());
+        if (ownConfiguration != configuration) {
+            throw new GradleException(
+                    String.format("""
+                        Cannot call getVersion with %s as it does not belong to project '%s'. getVersion can only \
+                        resolve a configuration that lives in the project it is called from. Either supply a \
+                        configuration from '%s', or call getVersion from the project that owns %s.\
+                        """, configuration, project.getPath(), project.getPath(), configuration));
+        }
```
felixdesouza, line 128: I think this might block upgrades potentially, I wonder whether it's worth it to gate it on erroring if we're on Gradle 9, and instead just log a warning for gradle 8 and below. It'd fail the gradle 9 upgrade PR and then something could be done about it.

src/test/java/com/palantir/gradle/versions/GetVersionPluginIntegrationTest.java
```diff
@@ -84,6 +84,19 @@ void get_version_is_callable_from_groovy_with_two_string_args(GradleInvoker grad
         InvocationResult result = gradle.withArgs("callGetVersion").buildsWithFailure();
         assertThat(result)
                 .output()
-                .contains("Unable to find 'com.google.guava:guava' in configuration ':unifiedClasspath'");
+                .contains("Unable to find 'com.google.guava:guava' in configuration ':gcvGetVersions'");
+    }
+
+    @Test
+    void get_version_fails_when_called_at_configuration_time(GradleInvoker gradle, RootProject project) {
```
felixdesouza, line 91: surprised we didn't have this test already

README.md
```diff
@@ -221,6 +221,9 @@ task printGuavaVersion {
 }
 ```
 
+> [!WARNING]
+> Any configuration you supply **must** live in the project that `getVersion` is being called from.
```
felixdesouza: See my above comment, re this is an error in Gradle 9, and we will error with a more helpful message if so etc.

src/main/java/com/palantir/gradle/versions/GetVersionPlugin.java
FinlayRJW: added a guard in 93c9fdef45c44b0c35a83568e44c9a5ca8079c17

README.md
FinlayRJW: addressed in 93c9fdef45c44b0c35a83568e44c9a5ca8079c17

src/test/java/com/palantir/gradle/versions/ConsistentVersionsPluginIntegrationTest.java
```diff
@@ -261,25 +265,47 @@ void get_version_function_works_for_a_dependency_not_declared_by_the_project(
     @DisabledConfigurationCache(
             "Cannot reference a Gradle script object from a Groovy closure as these are not supported with the"
                     + " configuration cache")
-    void get_version_function_fails_clearly_when_configuration_belongs_to_another_project(
-            GradleInvoker gradle, SubProject lib, SubProject consumer) {
+    @ParameterizedByGradleVersion(when = @WhenVersion(lessThan = "9.0", stringValue = "warn"), otherwiseString = "fail")
+    @AdditionallyRunWithGradle("9.3.0")
+    void get_version_function_rejects_a_configuration_belonging_to_another_project(
```
crogoz: nit: maybe `get_version_function_flags_a_configuration_belonging_to_another_project` ?

src/test/java/com/palantir/gradle/versions/ConsistentVersionsPluginIntegrationTest.java
```diff
@@ -261,25 +265,47 @@ void get_version_function_works_for_a_dependency_not_declared_by_the_project(
...
 
-        InvocationResult result =
-                gradle.withArgs(":consumer:demo", "--write-locks").buildsWithFailure();
-        assertThat(result)
-                .output()
-                .contains("does not belong to project ':consumer'")
-                .contains("getVersion can only resolve a configuration that lives in the project it is called from");
+        // Resolving another project's configuration breaks on Gradle 9, so getVersion fails hard there. On earlier
+        // versions it still works, so we only warn to avoid blocking consumers until they upgrade.
+        if (behaviourOnForeignConfiguration.equals("fail")) {
```
crogoz, line 296: I think I preferred how we used to check different behaviors in nebula-test for different gradle versions. 
I think it would be clearer here if we said/ inferred the version form ParameterizedByGradleVersion/inferred from gradleInvoker ?:

```
if (gradleVersion >= 9.x) {
 buildWithFailure...
} else {
buidsSuccessfully...
}
```

no action here, but curious what your thoughts are

FinlayRJW: yeah, this was built for cases like: https://github.com/palantir/witchcraft-java-logging/blob/bb3696b51491468ebac7c42f9979cb4d38dd55b1/gradle-witchcraft-logging/src/test/java/com/palantir/witchcraft/java/logging/gradle/testreport/TestReportFormattingPluginIntegrationTest.java#L123-L129 but probably doesn't quite fit here - I'll see if I can more directly check the gradle version else might be worth having a way to inject the gradle version?

FinlayRJW: hmm yeah dont think the framework exposes anything. I think we leave as is for now and can FLUP with a framework change if we want a way to more directly pass in the gradle version?

src/test/java/com/palantir/gradle/versions/ConsistentVersionsPluginIntegrationTest.java
FinlayRJW: 👍  - see [4efed5c](https://github.com/palantir/gradle-consistent-versions/pull/1667/commits/4efed5c077f111142f8dd1ae810a3b002014f2af)

crogoz (approved):

PR #1733, opened 2026-09-10
Exclude gcvGetVersions configuration in CheckUnusedConstraintsProjectPlugin

fkarlo:
## Before this PR

`writeResolvedCoordinatesTask` OOMs the gradle worker locally. 

`CheckUnusedConstraintsProjectPlugin` used to get the dependencies from `gcvGetVersions` in local projects, but that configuration's dependencies are explicitly the whole project again:
https://github.com/palantir/gradle-consistent-versions/blob/c4a1a78e018eb4fc14886f58f151e84eb5133cef/src/main/java/com/palantir/gradle/versions/GetVersionPlugin.java#L70-L74

 so each local project walks the root project's entire dependency tree.

## After this PR

==COMMIT_MSG==
Exclude gcvGetVersions configuration in CheckUnusedConstraintsProjectPlugin
==COMMIT_MSG==


FinlayRJW:
👍

PR #1734, opened 2026-09-11
Add test to validate that checkUnusedConstraints doesn't resolve unrelated dependencies

fkarlo:
## Before this PR

Follow up to https://github.com/palantir/gradle-consistent-versions/pull/1733 guarding against this case

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


FinlayRJW:
👍

